PR #62, opened 2016-05-11
Fix grammar in overview.md

jonathanp:
😄


allmarkedup:
My (German) wife is always correcting my use of it's/its!! Thanks :-)

Issue #123, opened 2016-08-16
Variants' contexts get extended from the default context.


matijs:
Variant contexts seem to be created by extending the default context. Consider the example below.

``` yaml

---
title: foo
context:
  options:
    - one
    - two
    - three
    - four
variants:
  -
    name: bar
    context:
      options:
        - alpha
        - beta
```

The options variable in the `bar` variant's context will contain `['alpha', 'beta', 'three', 'four']` which is not what I'd expect when setting `options` explicitly in the variant's own context. A workaround for now is to not define a context for the default version of a component.

allmarkedup:
Hi @matijs that is actually the expected behaviour. The top-level context object is inherited and merged into all variants. Although the array behaviour does seem to trip people up so it may be worth revisiting...!

You can set context data for the default variant _only_ by explicitly defining a variant with the name `default` (this normally happens behind-the-scenes) and then giving it it's own data, something like:

```yaml
---
title: foo
context:
  title: This is inherited unless overriden in a variant
variants:
-
  name: default
  context:
    options:
      - foo
      - bar
  -
    name: bar
    context:
      title: this overrides the default title
      options:
        - alpha
        - beta
```

matijs:
@allmarkedup the array behaviour was indeed what tripped me up. I did expect the merging, just not this deep :)

```yaml
name: default
label: not default 
```

Solves this (and more) though! Thanks.

allmarkedup:
Ok great. I think that the array behaviour (even though it was what I originally intended) is so counter-intuitive (based on people's feedback anyway!) that it should actually maybe considered a bug. I may well change it in a future release.

allmarkedup:
I've reviewed this behaviour and I'm re-opening this issue because I think the array merging behaviour should be considered a bug. The context object merging is intended but arrays should not be merged as it is definitely not what would be expected.

felixdorner:
I don't know if this comment fits here but it might be a point to consider:

When I try to place a component-A--variant in component-B the content set in its config gets replaced by the default content of component-A (same config). I know that I could use ‘render’ instead of ‘>’ to place components and keep the content but in my case, I’d like to change something minor and keep the content of component-A--variant while placing it in component-B without it falling back to the default content of component-A.

I hope my description is understandable. For me, the expected behavior is not met here.

Chapabu:
[...]

PR #135, opened 2016-08-29
Remove mergeProp's ability to merge arrays

matijs:
Solves #123
Update test accordingly

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -93,9 +93,7 @@ module.exports = {
     },
 
     mergeProp(prop, upstream) {
-        if (_.isArray(upstream)) {
-            return _.uniq(_.concat(upstream, _.castArray(prop)));
-        } else if (_.isPlainObject(upstream) && _.isPlainObject(prop)) {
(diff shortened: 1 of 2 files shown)


matijs:
In a rudimentary test this seems to solve the issue I had, I'm not sure if other things broke or if this is covered by tests.

allmarkedup:
Thanks @matijs - I'm actually just putting together a 1.1 release that should fix the array merging issue in the context data. However I'm still actually undecided if I want to change this behaviour in `mergeProp` itself because it is responsible for handling other situations as well in which the array merge may be the correct route to take. I'll leave this open until I make a firm decision but tbh I'll probably leave it as is for now.

matijs:
That's cool. Context data seems to be where this belongs anyway :)

dkhuntrods:
This is resolved by the upcoming 1.1 release - array merging issues are fixed within the context data.

sun:
Merging array subkeys made perfect sense for us. Would it make sense to turn this into a configuration setting instead of hardcoding one behavior or the other?

PR #180, opened 2016-11-24
Better BrowserSync support

tlenex:
* const `watchers` is reassigned on `destroy`, changed to let
* added browsersync error handler, and bs handler
* removed duplicated `browser` prop from browsersync defaults
* `proxy` default prop should be an object for better defaults merging
* added `server` default for more readability, note that you can define only one property `proxy` or `server`, defining both will throw an error.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ npm-debug.log
 /example/*
 /build/*
 /fractal.js
+.idea/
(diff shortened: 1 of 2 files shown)


tlenex:
Fractal is a great tool, much better than fabricator we used! 
It only needs more love from maintainers 😺

allmarkedup:
Thanks @tlenex - I'll check this out, much appreciated.

And I know I'm doing a terrible job of maintenance right now, but that is only due to time, not a lack of love, I assure you :-) Hopefully I can fix that soon and get some updates out for people!

PR #286, opened 2017-08-01
Add Utils internals package

dkhuntrods:
This adds the Utils package and tests.


README.md
```diff
@@ -25,14 +25,14 @@ We are moving development of Fractal v2.0 and all first-party add-ons into a 'mo
 #### Running locally
 
 1. Clone this repository
-2. Install dependencies - `npm install`
-3. Bootstrap the packages together using Lerna - `npm run bootstrap`
+2. Install dependencies - `yarn install`
```
allmarkedup: I actually left this as `npm` rather than `yarn` intentionally - the repo includes yarn as a dependency so that lerna and all npm scripts can use it, but the user doesn't need to have Yarn installed to run the tests or install the dependencies. Obviously they can if they want to but I felt like leaving the instructions referenceing `npm` was the most neutral approach. What do you think?

allmarkedup:
So looks like builds are failing because Travis doesn't know to run the `lerna bootstrap` step before running the tests - adding a `pretest` NPM script to run the bootstrap should work, i.e.

```js
"scripts": {
  "pretest": "yarn run bootstrap",
  // ...
}
```

README.md
dkhuntrods: Ok, I'd thought that might be the case, but it felt a bit weird, having read the Readme recommending yarn, and then running the scripts which were all yarn, but then having the npm ref - it looked a bit like an oversight. 
Maybe we could just update the readme to be more explicit about that?

allmarkedup: Tbh honest... the scripts don't actually have to use yarn either. Lerna is configured to use yarn via the lerna.json file. Everything else could be changed to plain `npm` as all non-lerna scripts are client-agnostic.

dkhuntrods: Ok, shall I swap out to npm for scripts then, and revert the Readme?
Re. the pretest script, only issue with that is that it would runs every time I ran the test locally as well. Can we create a separate test:ci script, or something to that effect?

allmarkedup: So..
1) yes to swapping out for npm in scripts
2) I hadn't considered that... I _think_ you can use the `script` property in the travis.yml to specify what npm script to use so could create a specific CI one: https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Using-Gulp

packages/internals/utils/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@frctl/utils",
+  "version": "2.0.0-beta.0",
+  "description": "Utilities and helpers for Fractal.",
+  "homepage": "http://fractal.build",
+  "repository": "https://github.com/frctl/fractal/tree/v2/packages/utils",
```
allmarkedup: Just noticed that the repo location is incorrect - should be `https://github.com/frctl/fractal/tree/v2/packages/internals/utils`

PR #287, opened 2017-08-02
Add console logging utility package

allmarkedup:
Provides a set of useful console log utilities for generating nice CLI output.


packages/internals/console/src/utils.test.js
```diff
@@ -0,0 +1,66 @@
...
+      expect(result.stack).to.equal(null);
+    });
+  });
+
+  describe('.prefix()', function () {
+    it('adds a prefix to the start of a string', function () {
+      expect(utils.prefix('abcdefg', '123 ')).to.equal('123 abcdefg');
+    });
+    it('prefixes the first non-empty line and indents the other lines to match', function () {
+      expect(utils.prefix('\n\nabcde\nfg\nhi', '!')).to.equal('\n\n!abcde\n fg\n hi');
```
dkhuntrods: Could this ever be confusing/unexpected?

allmarkedup: I don't think so - in fact I think the opposite is true, without it the behaviour for multiline strings feels wrong - i.e. the output would be:

```
PREFIX Lorem ipsum dolor sit amet, consectetur adipisicing elit.
Deserunt, quidem ducimus ab! Minima, commodi quos, modi debitis delectus quo suscipit voluptatibus, ipsam consectetur nam autem fugiat.
Autem debitis, ut possimus!
```

as opposed to:

```
PREFIX Lorem ipsum dolor sit amet, consectetur adipisicing elit.
       Deserunt, quidem ducimus ab! Minima, commodi quos, modi debitis delectus quo suscipit voluptatibus, ipsam consectetur nam autem fugiat.
       Autem debitis, ut possimus!
```

I can't think of any situation where the former would be preferred - and if so you can just append to the string directly instead of using the prefix helper. Would you agree?

allmarkedup: Unless there is a better word for 'prefix'?

dkhuntrods: With a name like prefix, I would expect the first TBH. I think the problem is just the name, as I can see what it's doing and why it's useful.
It's more akin to a hanging indent...

allmarkedup: Ok decision made to stick to `prefix` as no better names come to mind...

PR #288, opened 2017-08-03
Add custom Nunjucks environment for use in add-ons

allmarkedup:
This PR adds a new package to provide a customised, Fractal-specific Nunjucks environment for use in Fractal addons such as the inspector and static site builder.

### DIFF
--- a/packages/internals/funjucks/src/env.test.js
+++ b/packages/internals/funjucks/src/env.test.js
@@ -1,0 +1,41 @@
+/* eslint handle-callback-err: off */
+
+const nunjucks = require('nunjucks');
+const {expect} = require('../../../../test/helpers');
+const factory = require('./env');
+
+const env = makeEnv();
+
+describe('factory', function () {
+  it('exports a function', function () {
+    expect(factory).to.be.a('function');
+  });
+
+  describe('factory()', function () {
+    it('returns a fresh Nunjucks environment', function () {
+      expect(env).to.be.instanceOf(nunjucks.Environment);
+      expect(env).to.not.equal(makeEnv());
+    });
+
+    it('promisifies the `renderString` method of the returned environment', function () {
+      expect(env.renderString('foo')).to.be.instanceOf(Promise);
+    });
+
+    it('promisifies the `render` method of the returned environment', function () {
+      const result = env.render('foo.html');
+      expect(result).to.be.instanceOf(Promise);
+      result.catch(err => {});
+    });
+
+    it('leaves other methods untouched', function () {
+      expect(env.addFilter('foo', () => {})).to.not.be.instanceOf(Promise);
+    });
+
+    it('adds the fractal instance as a property on the environment');
+  });
+});
+
+function makeEnv() {
+  const fractal = {}; // TODO: Use proper Fractal instance
+  return factory(fractal);
+}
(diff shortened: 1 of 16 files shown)


packages/internals/funjucks/src/env.test.js
```diff
@@ -0,0 +1,47 @@
...
+    for (const filter of ['await', 'beautify', 'highlight', 'stringify']) {
+      it(`adds the ${filter} filter`, function () {
+        expect(() => env.getFilter(filter)).to.not.throw(Error);
+      });
+    }
+  });
+});
+
+function makeEnv() {
+  const fractal = {}; // TODO: Use proper Fractal instance
```
dkhuntrods: I'd say a mock Fractal instance would be better here, can use a proper instance in the integration tests.

packages/internals/funjucks/src/env.test.js
```diff
@@ -0,0 +1,47 @@
...
+      const result = env.render('foo.html');
+      expect(result).to.be.instanceOf(Promise);
+      result.catch(err => {});
+    });
+
+    it('leaves other methods untouched', function () {
+      expect(env.addFilter('foo', () => {})).to.not.be.instanceOf(Promise);
+    });
+
+    it('adds the fractal instance as a property on the environment');
```
dkhuntrods: Just checking this was intentionally left a stub.

packages/internals/funjucks/src/filters/await.test.js
```diff
@@ -0,0 +1,45 @@
...
+  it('exports a function', function () {
+    expect(factory).to.be.a('function');
+  });
+
+  it('returns an object with the expected properties', function () {
+    const filter = factory();
+    expect(filter.name).to.equal('await');
+    expect(filter.async).to.equal(true);
+    expect(typeof filter.filter).to.equal('function');
+  });
```
dkhuntrods: (Lines 5-14)
These look like they're pretty consistent across the filters, could be extracted to common filter tests?

packages/internals/funjucks/src/env.test.js
allmarkedup: Yeah i wasn't really sure of the best way to mock it (which is why the test above was also left as a stub). My intention was that when we have ported over the Fractal instance proper I can pull that in and do an `instanceof` check in the stubbed test. Is there a better way to handle this in the meantime?

packages/internals/funjucks/src/filters/await.test.js
allmarkedup: Yes definitly, will be the same for the plugins, but I'd like to chat with you first to make sure we have a good re-usable pattern for it across the monorepo. Unless you have any objections I'll merge it in like this and we can come back to it once we have an good approach figured out.

packages/internals/funjucks/src/env.test.js
dkhuntrods: If there's direct functionality that the testee might require, a mock or stub with Sinon that would mock the functionality; if it's simply to check that that thing *is a* Fractal, and that's all, then what you've outlined is probably the simplest for now.
There are other options: https://github.com/sinonjs/sinon/issues/567#issuecomment-58648756

PR #289, opened 2017-08-04
V2 internals supports

dkhuntrods:
An initial PR for the support package: as Collection is quite large, doing this a separate first PR and will include the other portions in a later PR(s) to keep the size down.


packages/internals/utils/src/utils.js
```diff
@@ -23,9 +23,6 @@ const utils = module.exports = {
         if (!_.isPlainObject(sourceValue) || !_.isPlainObject(objectValue)) {
           return sourceValue;
         }
-        if (_.isUndefined(sourceValue)) {
-          return objectValue;
-        }
```
allmarkedup: What's the reason behind removing this? As far as I remember it was needed to ensure that an undefined value in the source object has it's value correctly filled by the equivalent property value in the defaults object. Or am I missing something?

allmarkedup: As per slack convo looks like this was never needed... win 👍

PR #291, opened 2017-08-04
Add configuration manager package

allmarkedup:
This PR adds a configuration manager class for storing, validating and retrieving configuration properties.

The class also provides support for defining 'accessors' that operate on property values at the point of retrieval and so provide a hook by which to sanitise data before it is returned.


packages/internals/config/src/accessors/package-loader.test.js
```diff
@@ -0,0 +1,46 @@
...
+  });
+
+  it('loads packages from the cwd if the target is a path or an array with a path as it\'s first item', function () {
+    const pkg = loadPkg('./test/fixtures/add-ons/plugin');
+    expect(pkg).to.be.an('object');
+    expect(pkg).to.have.property('name');
+
+    const pkg2 = loadPkg(['./test/fixtures/add-ons/plugin']);
+    expect(pkg2).to.be.an('object');
+    expect(pkg2).to.have.property('name');
```
dkhuntrods: Might be worth testing the value of the name explicitly.

packages/internals/config/src/accessors/package-loader.test.js
```diff
@@ -0,0 +1,46 @@
...
+  });
+
+  it('throws an error if a package is not found', function () {
+    expect(() => loadPkg('./not/found')).to.throw();
```
dkhuntrods: Should this have an error type and code?

packages/internals/config/src/config.test.js
```diff
@@ -0,0 +1,222 @@
...
+    it('runs the result through all relevant accessors', function () {
+      const config = new Config({
+        data: {
+          foo: {
+            bar: 'baz'
+          }
+        }
+      });
+      config.addAccessor('foo.bar', value => '!' + value);
+      config.addAccessor('foo', value => '@' + value);
```
dkhuntrods: Looking at the data object, I would have expected config.get('foo') to return the object, so wondering how this combination is calculated.

packages/internals/config/src/config.test.js
```diff
@@ -0,0 +1,222 @@
...
+  describe('.getAccessorsForPath()', function () {
+    it('retrieves all accessors that match the dot-notation path', function () {
+      const config = new Config();
+      config.addAccessor('foo.bar', val => val);
+      config.addAccessor('foo.bar.baz', val => val);
+      config.addAccessor('bax.foo.bar', val => val);
+      config.addAccessor('fooooo', val => val);
+      config.addAccessor('bar', val => val);
+      expect(config.getAccessorsForPath('foo.bar').length).to.equal(1);
+      expect(config.getAccessorsForPath('foo.bar.baz').length).to.equal(2);
```
dkhuntrods: Why is this '2' - is it because of the preceding accessor? If so, not sure I would've expected that.

PR #292, opened 2017-08-05
Add alias-able module loader class

allmarkedup:
A small wrapper class for Webpack's `enhanced-resolve` that provides support for requiring and resolving modules against a set of path aliases. It will be used to enable path-independent requiring of local modules in Fractal config files (for example, shared data files).

### DIFF
--- a/packages/internals/loader/src/loader.test.js
+++ b/packages/internals/loader/src/loader.test.js
@@ -1,0 +1,74 @@
+const {join} = require('path');
+const Resolver = require('enhanced-resolve/lib/Resolver');
+const {ResolverFactory} = require('enhanced-resolve');
+const {expect, sinon} = require('../../../../test/helpers');
+const Loader = require('./loader');
+
+describe('Loader', function () {
+  describe('constructor()', function () {
+    it('passes configuration options to the resolver', function () {
+      const spy = sinon.spy(ResolverFactory, 'createResolver');
+      // eslint-disable-next-line no-unused-vars
+      const loader = new Loader({
+        fileSystem: 'foo'
+      });
+      expect(spy.args[0][0].fileSystem).to.equal('foo');
+      spy.restore();
+    });
+  });
+
+  describe('.resolve()', function () {
+    it('synchronously resolves the path with respect to the root', function () {
+      const loader = new Loader();
+      expect(loader.resolve('./loader', __dirname)).to.equal(join(__dirname, 'loader.js'));
+    });
+
+    it('resolved paths relative to the cwd if no root is supplied', function () {
+      const loader = new Loader();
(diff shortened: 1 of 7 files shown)


packages/internals/loader/src/loader.test.js
```diff
@@ -0,0 +1,74 @@
...
+    it('requires a module after resolving the path', function () {
+      const loader = new Loader();
+      const spy = sinon.spy(loader, 'resolve');
+      loader.require('./loader', __dirname);
+      expect(spy.calledWith('./loader', __dirname)).to.equal(true);
+    });
+
+    it('throws an error if the module cannot be required', function () {
+      const loader = new Loader();
+      expect(() => loader.require('./foo', __dirname)).to.throw();
```
dkhuntrods: Worth adding/wrapping our own error message here? I'd say this could be a fairly common error for people to encounter.

packages/internals/loader/src/loader.test.js
```diff
@@ -0,0 +1,74 @@
...
+
+    it('resolved paths relative to the cwd if no root is supplied', function () {
+      const loader = new Loader();
+      const result = loader.resolve('./test/fixtures/add-ons/plugin');
+      expect(result).to.equal(join(process.cwd(), 'test/fixtures/add-ons/plugin.js'));
+    });
+
+    it('throws an error if the path cannot be resolved', function () {
+      const loader = new Loader();
+      expect(() => loader.resolve('./foo', __dirname)).to.throw();
```
dkhuntrods: Worth adding/wrapping our own error message here? I'd say this could be a fairly common error for people to encounter.

packages/internals/loader/src/loader.test.js
allmarkedup: Would the idea behind that be just so we can test the error situation better? We'll have to grab the error message out of the original error and relay that anyway so I thought it maybe it should just be the standard Node failed `module.require` error here rather than a custom one. But happy to add one in if there is a good reason for it.

dkhuntrods: Not mainly for testing - as we'd be altering the standard `module.require` methods, I feel it would be helpful to give people more context around the standard error when it occurs (I think less experienced users might not realise that it's using `require` behind the scenes, or how these classes change that behaviour, and may find it tricky to debug)

allmarkedup: Ok sure, makes sense. So catch the error, amend the message and re-throw - does that sound good?

dkhuntrods: Yep!

dkhuntrods: Should add here, look at integrating something like https://github.com/joyent/node-verror

PR #293, opened 2017-08-07
V2 internals supports

dkhuntrods:
Migrates Component, ComponentCollection, and Entity classes, and tests for the index.
Adds skeleton support for File and FileCollection.


allmarkedup:
Looking great! I've had a few thoughts about component instance instantiation and responsibilities but nothing that should change the user-facing API or conflict with the work done here so will save that for a separate discussion and possible additional PR later. Great job :-)

PR #298, opened 2017-08-08
Improve and simplify Component class construction

dkhuntrods:
This eliminates redundant properties and sets defaults if not provided.
Additionally, 
- add better validation for Collection classes, and 
- add [Symbol.toStringTag] references for testing and stack traces

### DIFF
--- a/packages/internals/support/src/component.js
+++ b/packages/internals/support/src/component.js
@@ -27,6 +27,10 @@ class Component extends Entity {
     super(props);
   }
 
+  get [Symbol.toStringTag]() {
+    return 'Component';
+  }
+
   static isComponent(item) {
     return item instanceof Component;
   }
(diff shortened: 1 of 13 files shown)


packages/internals/support/src/component.js
```diff
@@ -1,30 +1,73 @@
...
 const File = require('./file');
 const FileCollection = require('./file-collection');
 
+const isFile = value => value instanceof File;
+check.mixin(isFile, 'file');
+const isCollection = value => value instanceof Collection;
+check.mixin(isCollection, 'collection');
+const isFileCollection = value => value instanceof FileCollection;
+check.mixin(isFileCollection, 'fileCollection');
+
```
allmarkedup: I like this... could be useful to have our own pre-customised `check` instance which has these Fractal specific tests already mixed in as a shared resource somewhere. Maybe as part or utils/supports? A task for later maybe...

allmarkedup:
All looks spot on. Merging!

packages/internals/support/src/component.js
dkhuntrods: Yep, was thinking the same.

PR #299, opened 2017-08-09
Console simplification/improvements

allmarkedup:
Now uses allmarkedup/climate for string rendering, simplifies the exports and generally cleans things up a lot :-)


allmarkedup:
@dkhuntrods of course!!! Who do you think I am.... ! ;-)

PR #300, opened 2017-08-09
Begin implementation of Parser: Add Plugins and Files transform

dkhuntrods:


packages/internals/parser/src/transform/plugins.js
```diff
@@ -0,0 +1,57 @@
...
+    if (items) {
+      this.add(items);
+    }
+  }
+
+  /**
+   * Adds a new item to the store
+   *
+   * @param  {object|array} item Item or array of items to add
+   * @return {Store} Returns a reference to itself
```
allmarkedup: Minor point but `@return` value should be {Plugins} not {Store} here

packages/internals/parser/src/transform/plugins.js
```diff
@@ -0,0 +1,57 @@
...
+const isPlugin = check.schema.bind(null, pluginSchema);
+
+const _items = new WeakMap();
+const _opts = new WeakMap();
+
+class Plugins {
+  constructor(items, config = {}) {
+    _items.set(this, []);
+
+    // QUESTION: options are init-only as making it changeable
```
allmarkedup: Yeah definitely should be init-only!

packages/internals/parser/src/transform/plugins.js
dkhuntrods: Thanks for catching!

PR #304, opened 2017-08-10
A few tweaks

allmarkedup:
Just a few minor tweaks now that we can start hooking a few things together...


allmarkedup:
@dkhuntrods Might actually want to un-Emitterify the core fractal instance if we are going to move to the event-emitting-promise model as per out discussion... have left it for now until we've proved it out but just mentioning here as a reminder for later.

PR #308, opened 2017-08-11
Review proposed version of EmittingPromise

dkhuntrods:
An implementation of the EmittingPromise, as discussed. 
**This PR can be for review purposes only.**

### DIFF
--- a/packages/internals/support/src/emitting-promise.js
+++ b/packages/internals/support/src/emitting-promise.js
@@ -1,10 +1,13 @@
+/* eslint guard-for-in: "off" */
 const {EventEmitter2} = require('eventemitter2');
 const assert = require('check-types').assert;
 
 class EmittingPromise extends Promise {
   constructor(resolver) {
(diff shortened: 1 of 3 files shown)


packages/internals/support/src/emitting-promise.js
```diff
@@ -0,0 +1,30 @@
+const {EventEmitter2} = require('eventemitter2');
+const assert = require('check-types').assert;
+
+class EmittingPromise extends Promise {
+  constructor(resolver) {
+    assert.function(resolver, `EmittingPromise constructor requires a resolver with params 'resolve, reject [, emit]' [invalid-resolver]`);
+    const emitter = new EventEmitter2();
```
allmarkedup: If we want to support wildcards you need to explicitly enable it in EE2 via `{wildcard: true}` ... not sure why it's not the default behaviour tbh! :-)

dkhuntrods: Thanks for the heads up!

PR #311, opened 2017-08-13
Add basic renderer package

allmarkedup:
Note that `renderComponent` and `renderVariant` methods are just stubs right now. Also hooks the renderer up to the main Fractal instance.

### DIFF
--- a/packages/internals/renderer/src/adapter-store.js
+++ b/packages/internals/renderer/src/adapter-store.js
@@ -2,7 +2,7 @@ const {remove} = require('lodash');
 const {File} = require('@frctl/support');
 const {toArray} = require('@frctl/utils');
 const debug = require('debug')('fractal:renderer:adapters');
-const adapter = require('./adapter');
+const Adapter = require('./adapter');
 
 const _adapters = new WeakMap();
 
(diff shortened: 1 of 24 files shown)


packages/internals/renderer/src/adapter-store.js
```diff
@@ -0,0 +1,51 @@
...
+    for (const adapter of _adapters.get(this)) {
+      if (adapter.match(file)) {
+        return adapter;
+      }
+    }
+  }
+
+  add(items) {
+    const adapters = _adapters.get(this);
+    toArray(items).map(props => new Adapter(props)).forEach(adapter => {
```
dkhuntrods: I know the Adapter will error here it `props` are invalid, but it might be worth wrapping the error at this level?

packages/internals/renderer/src/adapter.js
```diff
@@ -0,0 +1,40 @@
...
+    Validator.assertValid(props, adapterSchema, 'Adapter schema invalid [adapter-invalid]: ');
+    _props.set(this, props);
+  }
+
+  match(file) {
+    if (!File.isFile(file)) {
+      throw new TypeError(`Adapter.match - file argument must be an instance of File [file-invalid]`);
+    }
+    const props = _props.get(this);
+    const match = isFunction(props.match) ? props.match : toArray(props.match).map(ext => normalizeExt(ext));
```
dkhuntrods: Does this conversion need to happen every time `match` is called? Could it be done once on init?

allmarkedup: Yes that would be better! It was actually that way before I converted it to class but slipped in there somehow. Will move it to the constructor.

packages/internals/renderer/src/adapter-store.js
allmarkedup: Tbh I'm not sure what we'd gain from that here... the error message thrown by the adapter is really the one that is useful (as it contains the validation errors) and seems a little overkill to catch and throw it again here.  I think I will prob leave like this for now and we can always revisit if it's confusing later.

PR #315, opened 2017-08-14
Reorganise support package

allmarkedup:
Organises the support folder by type and also moves in add-on classes and schemas that were previously part of other packages. 

Also @dkhuntrods you were right, the Renderer class depending on Fractal was a bad smell and actually totally unnecessary. I've unpicked it now so it doesn't even require the config instance. So maybe for consistency, for the Parser let's just pass a plain object of opts into the constructor, not a Config instance. I.e:

```js
const parser = new Parser({
  src: [...],
  transforms: [...],
  plugins: [...],
})
```

Hopefully that makes sense :-)


packages/internals/renderer/src/renderer.test.js
```diff
@@ -24,29 +24,20 @@ const component = new Component({
...
+  return new Renderer(toArray(adapter));
 }
 
 describe('Renderer', function () {
   describe('constructor', function () {
-    it('throws an error if no Fractal instance is provided', function () {
-      expect(() => new Renderer('fractal')).to.throw('[fractal-required]');
+    it('throws an error on invalid arguments', function () {
+      expect(() => new Renderer('foo')).to.throw('[adapters-invalid]');
       expect(() => makeRenderer()).to.not.throw('[fractal-required]');
```
dkhuntrods: This probably doesn't need the string param, just `not.throw()` should be fine

packages/internals/support/src/add-ons/command.test.js
```diff
@@ -0,0 +1,157 @@
...
+  });
+
+  describe('.handler', function () {
+    it('returns the command handler', function () {
+      const cmd = command(validCommand);
+      expect(cmd.handler).to.equal(validCommand.handler);
+    });
+  });
+
+  // describe.skip('.handler()', function () {
```
dkhuntrods: Are these no longer relevant, or waiting on something else?

allmarkedup: No longer relevant, that logic has now been moved to the CLI package as it was requiring too much weird wrangling of scope/args.

packages/internals/renderer/src/renderer.test.js
allmarkedup: Very true, copy and paste coding at it's best :-)

packages/internals/support/src/add-ons/command.test.js
dkhuntrods: Ok, worth deleting then?

allmarkedup: Done!

PR #316, opened 2017-08-15
EmittingPromise: Pass entire emitter object to resolver

allmarkedup:
@dkhuntrods so I think this works?? Test all pass and I can't think of any negative side affects. What do you think?

### DIFF
--- a/packages/internals/support/src/emitting-promise.js
+++ b/packages/internals/support/src/emitting-promise.js
@@ -9,13 +9,13 @@ class EmittingPromise extends Promise {
       wildcard: true
     });
 
-    super((resolve, reject) => {
-      resolver(resolve, reject, (eventName, ...args) => {
-        process.nextTick(() => {
-          emitter.emit(eventName, ...args);
-        });
-      });
-    });
+    const oldEmit = emitter.emit;
+    emitter.emit = function (...args) {
+      process.nextTick(() => oldEmit.bind(emitter)(...args));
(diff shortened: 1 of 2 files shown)


packages/internals/support/src/emitting-promise.js
```diff
@@ -9,13 +9,13 @@ class EmittingPromise extends Promise {
...
     });
 
-    super((resolve, reject) => {
-      resolver(resolve, reject, (eventName, ...args) => {
-        process.nextTick(() => {
-          emitter.emit(eventName, ...args);
-        });
-      });
-    });
+    const oldEmit = emitter.emit;
```
dkhuntrods: Could the function be bound once here, rather than in every call to `emit`?

allmarkedup:
R.e. tests 'really' passing - yes that was a concern to me too but they are! Covering that is why I put in the 'counter' check for the number of `expects` around line 64. I'm sure an explicit `expects` counter for Chai probably exists as a plugin? Might be useful to integrate at some point.

allmarkedup:
Oops didn't mean to close! :-)

packages/internals/support/src/emitting-promise.js
allmarkedup: Yes definitely! Good spot :-)

allmarkedup:
@dkhuntrods and yes feel free to update and merge!

PR #318, opened 2017-08-16
Grab bag: simplify renderer, add watch/unwatch methods, delete code

allmarkedup:
Because less code is better, right? ;-)


allmarkedup:
@dkhuntrods I'm going to merge this in so I can branch off `v2` for some other stuff. Hope that's ok! If you spot anything dodgy looking I'll fix it later ;-)

packages/internals/config/src/config.test.js
```diff
@@ -211,6 +211,29 @@ describe('Config', function () {
...
+  describe('.pick()', function () {
+    it('returns an object consisting of key/values picked from the data', function () {
+      const config = new Config({
+        foo: 'bar',
+        one: 'two',
+        nested: {
+          prop: 'child'
+        }
+      });
+      expect(config.pick('foo', 'nested.prop')).to.eql({
```
dkhuntrods: I would expect 'nested.prop' to equal 'child', or have I missed something?

PR #324, opened 2017-08-22
Complete implementation of fractal-transform-components module

dkhuntrods:
Does what it says on the tin


allmarkedup:
@dkhuntrods Looks good! Will merge it in and then we can add the `collection` name property we just discussed separately once that is implemented.

PR #326, opened 2017-08-22
Extract shared app code into standalone package

allmarkedup:
@dkhuntrods hooking all this stuff up it really seemed like `plugin.transform` should just be a required property now. So I made it that (and updated the parser accordingly). So now no need for the concept of a 'default' transform - you either add a transform property or it'll shout at you, which I think should be both easier to document and also simpler!


packages/internals/app/src/app.test.js
```diff
@@ -0,0 +1,271 @@
...
+    it('accepts configuration data', () => {
+      const app = makeApp();
+      expect(app.config.data).to.eql(config);
+    });
+
+    it('does not throw an error if no config data is provided', () => {
+      expect(() => new App()).to.not.throw();
+    });
+
+    it('does not throw an error if no config data is provided', () => {
```
dkhuntrods: Duplicate here

PR #327, opened 2017-08-23
WIP/Review only

dkhuntrods:
This is a WIP branch, PR for review purposes only

### DIFF
--- a/packages/internals/support/src/entities/component.js
+++ b/packages/internals/support/src/entities/component.js
@@ -1,46 +1,33 @@
-const assert = require('check-types').assert;
+const {assert} = require('check-types');
 const check = require('check-more-types');
 const {omit} = require('lodash');
+const {cloneDeep, isObjectLike, get, set} = require('lodash');
 const Collection = require('../collections/collection');
 const FileCollection = require('../collections/file-collection');
 const Entity = require('./entity');
 const File = require('./file');
 
-const isFile = value => value instanceof File;
-check.mixin(isFile, 'file');
-const isCollection = value => value instanceof Collection;
-check.mixin(isCollection, 'collection');
-const isFileCollection = value => value instanceof FileCollection;
-check.mixin(isFileCollection, 'fileCollection');
-
-const componentSchema = {
-  path: check.maybe.unemptyString,
-  relative: check.maybe.unemptyString,
-  src: check.file,
-  name: check.maybe.unemptyString,
-  config: check.maybe.object,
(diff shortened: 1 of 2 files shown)


packages/internals/support/src/entities/component.js
```diff
@@ -64,6 +51,25 @@ class Component extends Entity {
...
+  get(path, fallback) {
+    fallback = get(this.config, path, fallback);
+    const result = get(_data.get(this), path, fallback);
+    return isObjectLike(result) ? cloneDeep(result) : result;
+  }
+
+  set(path, value) {
+    assert.string(path, 'Component.set - `path` argument must be a string [path-invalid]');
+    // TODO: extract shared caching code with config: this.removeFromCache(path);
+    set(_data.get(this), path, cloneDeep(value));
```
allmarkedup: Have you checked this with non-array/object values? I think i you try and do cloneDeep on strings it might behave oddly (like return an array) which is why i used `isObjectLike(result) ? cloneDeep(result) : result;` in the getter... may have been me just being an idiot but worth double checking.

allmarkedup: Also... just thinking... do we need to cloneDeep on `get` if we are doing it here too?

dkhuntrods: The tests do contain string values.
I'm pretty sure it checks that itself (though I'd also copied that check in the getter, I should at least be consistent!)
I've also checked it out on runkit, seems to just pass primitives back - though it just seems to have gone down. I think I broke it 😇

dkhuntrods: I believe we do, we'd want to break the link to any assigned objects both coming and going, in case something else were to modify the original passed-in value after being either assigned or returned.

packages/internals/support/src/entities/component.js
```diff
@@ -1,71 +1,112 @@
...
+    Object.defineProperty(core, 'path', {
       enumerable: true,
-      get: pathGet
+      get: () => {
+        return get(_defaults.get(this), 'src').path;
+      }
     });
-
-    Object.defineProperty(this, 'relative', {
+    Object.defineProperty(core, 'relative', {
```
allmarkedup: Definitely a lot cleaner defining getters on the core data object like this than having `if (path = 'relative')` checks in there! 👍

dkhuntrods: Yes! Had thought it was worth a try and it did the trick

PR #328, opened 2017-08-23
[WIP/spike] - Sketch out suggestions on component property handling

allmarkedup:
@dkhuntrods so I had a bit of a think about defaults, required props/items, config loading etc... sketched out (basically pseudo-code right now) some thoughts which we should discuss when we can. Hopefully the amends give you a feel for some of my thinking!

I'm definitely leaning towards this `.get/.set` and proxied properties approach though... I think there are some interesting things this unlocks and was getting quite excited playing with it ;-)

I've got a list of other edge-cases/talking points but I'll save listing them out until we can chat about it.


allmarkedup:
Oh and I didn't try and write or update any tests on this.... didn't seem worth it until it's been discussed in a bit more detail.

PR #333, opened 2017-08-30
Refactor of Entity to use shadowing on config properties

dkhuntrods:
Refactor of Entity, Components et al to use shadowing on config properties.
Also adds various initialisation conveniences on Collections, and some new Collection classes

### DIFF
--- a/packages/internals/support/src/entities/component.js
+++ b/packages/internals/support/src/entities/component.js
@@ -1,73 +1,114 @@
-const assert = require('check-types').assert;
-const check = require('check-more-types');
-const {omit} = require('lodash');
-const Collection = require('../collections/collection');
+const {normalizeName, uniqueName} = require('@frctl/utils');
+const Validator = require('../validator');
+const schema = require('../../schema');
+const VariantCollection = require('../collections/variant-collection');
 const FileCollection = require('../collections/file-collection');
 const Entity = require('./entity');
-const File = require('./file');
-
-const isFile = value => value instanceof File;
-check.mixin(isFile, 'file');
-const isCollection = value => value instanceof Collection;
-check.mixin(isCollection, 'collection');
-const isFileCollection = value => value instanceof FileCollection;
-check.mixin(isFileCollection, 'fileCollection');
-
-const componentSchema = {
-  path: check.maybe.unemptyString,
-  relative: check.maybe.unemptyString,
-  src: check.file,
-  name: check.maybe.unemptyString,
-  config: check.maybe.object,
-  variants: check.maybe.collection,
-  files: check.maybe.fileCollection
-};
+const Variant = require('./variant');
 
 const _src = new WeakMap();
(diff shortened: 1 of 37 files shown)


packages/internals/support/src/entities/component.js
```diff
@@ -1,101 +1,120 @@
...
+
+  getVariants() {
+    return _variants.get(this);
+  }
+
+  getVariant(name) {
+    return _variants.get(this).find(name);
+  }
+
+  addVariant(variant) {
```
allmarkedup: I think that people should be able to supply a plain object here and have the component create the Variant instance from it (so that they don't need to add the component name property, generate a name etc). Actually I'd say that maybe that functionality could live on the VariantCollection itself (i.e. pushing/concating an object would auto-instantiate a Variant) but that could open up a whole can of worms with regards to the other collections....! Maybe just have the component create the variant and then add it to the VariantCollection for now?

packages/internals/support/src/schema/component.schema.js
```diff
@@ -0,0 +1,27 @@
...
+module.exports = {
+  $schema: 'http://json-schema.org/draft-06/schema#',
+  definitions: {
+    file: fileSchema,
+    fileCollection: fileCollectionSchema
+  },
+  type: 'object',
+  properties: {
+    src: {
+      $ref: '#/definitions/file'
```
allmarkedup: So I'm not sure referencing a File schema here is correct - it actually needs to be a `File` instance (as instantiating a component with a plain object that matches the file schema will still cause errors). I added custom `instanceof` checks to the validator on my pages branch last night so that stuff like this works:
```
src: {
   instanceof: 'File'
}
```

but I've not merged that in to V2 yet. Maybe let's leave as is for now and update when I've merged in the Validator updates?

packages/internals/support/src/entities/component.js
```diff
@@ -1,101 +1,120 @@
...
+    _variants.set(this, _variants.get(this).push(variant));
+  }
+
+  getViews() {
```
allmarkedup: Let's merge it in without these *View methods for now... I need to look over the current prototype views stuff again and then I can add these in if we feel they are needed.

packages/internals/support/src/entities/component.js
dkhuntrods: Yes, I was wondering about that. Sounds good. 👍

packages/internals/support/src/schema/component.schema.js
dkhuntrods: In this instance, the file schema will match either a File instance or file props, but a clear `instanceof` is certainly cleaner in general!

allmarkedup: Yeah the problem is that it _would_ match file props and so validate it as being ok - but it wouldn't actually work as the component requires an actual File instance

Issue #160, opened 2016-10-06
fractal start --sync does not load local server


DadaConsultant:
On windows environment - when you set up fractal using the guide provided on the site. Running fractal start --sync only try to open fractal.js but fails to launch the web server UI.

DadaConsultant:
Ever mind, resolved the issue. I wasn't running the command from the 'public/' folder created in my root dir

allmarkedup:
@DadaConsultant Glad you got it working - however it shouldn't actually need to be run from the public directory, it should work from the project root directory, so that may be a bug - I'll look into it.

schnubb:
It mustn't be the public directory, but on cmd you have to start fractal from a subdirectory.
Anything new on this bug?

siiron:
Having the same issue on Windows 10 running node v6.9.5 and fractal v1.1.3. Any progress investigating fixing this?

mja202:
Having the same problem on Mac. Work around for me is to start the Fractal CLI and then start the server from there.

Using yarn:
`yarn run fractal`

Within fractal:
` start --sync`

serslon:
I fixed this problem after run command 
`npm install vorpal@1.11.4`

But I have fractal version 1.0.8

dkhuntrods:
@mja202 @serslon There was an issue with an updated version of vorpal, see this thread for more details: #232

trickydisco78:
I#'m having the same issue on windows.

Jsut tried the same as mj202 recommended and it worked!

julmot:
Just went into this issue too. Fractal seems to be a great idea, but it needs more love. The approach of @mja202 isn't a solution, just a workaround.

julmot:
Btw: What you can do to speed it up in one command is: `yarn run fractal -- start --sync`

mja202:
Yes, it was a workaround. However, the current fractal (@1.1.4) works as it should (at least for me).

When fractal is installed globally (`yarn global add @frctl/fractal`) then `fractal start sync` within a project should just work.

allmarkedup:
@julmot @trickydisco78 can you confirm that you are using the latest release (`1.1.4`)? This was caused by a breaking change in the version of Vorpal that was used but it was bumped in the latest release and should work now.

julmot:
@allmarkedup I was using 1.1.4, yes.

julmot:
I was able to get it to work with [this workaround](https://github.com/frctl/fractal/issues/118#issuecomment-255254117).

[...]

PR #348, opened 2017-09-18
Use new mode API property. Closes #346

### LINKED ISSUE #346: Greenkeeper update broke fang API use
I might be wrong about this, but I'm having an issue whereby the output of fang@2.x is inconsistent with the output of fang@1.x, which the code in the config loader relies on.

The config loader relies on a key `mode` coming back from fang (see https://github.com/frctl/fractal/blob/master/src/core/data.js#L50), but fang no longer returns a `mode` key, instead going for two possible separate `*_mode` keys: `ace_mode` and `codemirror_mode`.

As such, either the `fang` dep should be reverted to `^1.0.0` or the code should be updated to use the `ace_mode` or `codemirror_mode` key (though it's worth saying that different languages in `fang` appear to sometimes not have the `codemirror_mode` key in v2, `ace_mode` appears to be consistent, though).

julmot:
This fixes the error

```bash
Error loading data file ....config.yml: Cannot read property 'toLowerCase' of undefined
```

in the `$ fractal build` command. Even though I'm still experiencing [this issue](https://github.com/frctl/fractal/issues/160#issuecomment-330337482) when using `$ fractal start --sync`.

### DIFF
--- a/src/core/data.js
+++ b/src/core/data.js
@@ -74,7 +74,7 @@ module.exports = {
 
     writeFile(filePath, data) {
         const pathInfo = Path.parse(Path.resolve(filePath));
-        const format = utils.lang(filePath, true).mode;
+        const format = utils.lang(filePath, true).ace_mode;
         return fs.writeFileAsync(filePath, this.stringify(data, format));
     },
 
--- a/src/core/data.js
+++ b/src/core/data.js
@@ -47,7 +47,7 @@ module.exports = {
     },
 
     readFile(filePath) {
-        const format = utils.lang(filePath, true).mode;
+        const format = utils.lang(filePath, true).ace_mode;
         if (format === 'js' || format === 'javascript') {
             try {
                 filePath = Path.relative(__dirname, filePath);


src/core/data.js
```diff
@@ -47,7 +47,7 @@ module.exports = {
     },
 
     readFile(filePath) {
-        const format = utils.lang(filePath, true).mode;
+        const format = utils.lang(filePath, true).ace_mode;
```
ten1seven: I noticed there's also a reference to `utils.lang(filePath, true).mode` on [line 77](https://github.com/frctl/fractal/blob/6831f8000a5ffb9f8ba7d0977f9320dc0d198f49/src/core/data.js#L77).

julmot: You're right, I don't know where it's used but I missed that. Going to fix this right now.

julmot: Fixed. @ten1seven Btw.: Are you experiencing [this issue](https://github.com/frctl/fractal/issues/160#issuecomment-330337482) too?

ten1seven:
@julmot even with this PR, are you able to get localhost working? I took the other route and did a find/replace for `ace_mode` > `mode` which cleared up errors, but leaves localhost still failing.

julmot:
@ten1seven As mentioned, this fixes `$ fractal build` for me, but `$ fractal start --sync` is still failing without any error. So I'm experiencing the same issue like you. I don't know if this is related to this issue without digging into the code.

julmot:
@ten1seven Regarding [this issue](https://github.com/frctl/fractal/issues/160#issuecomment-330337482): I've now digged into it a bit and couldn't find any problems in the debug log. The reason is an unverified dependency update on Aug 11 that was submitted as PR from Greenkeeper and was merged without a manual test (I guess):

2017-09-18 22_49_27-history for package json - frctl_fractal

I've restored old dependencies and pushed them here: https://github.com/julmot/fractal/tree/debug

You can change your package.json dependency to the following as intermediate solution until an official fix was submitted:

```js
"@frctl/fractal": "julmot/fractal#debug",
```

ten1seven:
That's a lifesaver, thanks!

julmot:
I'm having something like this in my preview template:

```html
<link rel="stylesheet" type="text/css" href="{{path "/development/development.css"}}">
```

When exported using `$ npm run build ` I've also noticed that the path is invalid now. @ten1seven Can you reproduce this too?

ten1seven:
I'm not seeing that issue, but highlight.js seems to be failing.

2017-09-18 06 02 28 pm

alexrussell:
@julmot you should also change the non-fang output of `utils.js` to use `ace_mode` - see https://github.com/frctl/fractal/blob/master/src/core/utils.js#L17

sun:
[fractal v1.1.7](https://github.com/frctl/fractal/commits/v1.1.7) reverted dependencies in https://github.com/frctl/fractal/commit/197ff5b3c57e19472c2186a0da04e8f78d58d4fc including fang, which fixed #346 and also made this PR here obsolete – unless an upgrade to fang v2 is on the agenda?

julmot:
A dependency revert can only be an intermediate solution @sun

dkhuntrods:
@julmot Given that we're not actively developing v1 any more -- our efforts are focussed on getting v2 out as quickly as possible -- we're not planning on updating those dependencies in the near future.

PR #354, opened 2017-09-19
V2 Pages prep

allmarkedup:
This is a horribly messy PR that includes a bunch of Pages preparatory work in addition to some initial Pages implementation that I subsequently removed in order to get the prep work merged in..... and I can't figure out how to clean it up.

@dkhuntrods Perhaps we can just not look too closely at the commit history and just get it merged in so that we can pretend this ever happened? ;-)


packages/fractal/fractal/src/fractal.test.js
```diff
@@ -123,15 +121,16 @@ describe('Fractal', function () {
       const fractal = makeFractal();
```
dkhuntrods: I can't leave a comment on the exact line as there were no changes, but this applies to `'Can render components'` - as discussed, we should just test the component-rendering aspect, so the collection should be mocked via `getParser`

packages/fractal/fractal/src/fractal.test.js
```diff
@@ -123,15 +121,16 @@ describe('Fractal', function () {
...
       const renderer = new Renderer(fractal.get('adapters'));
       const spy = sinon.spy(renderer, 'render');
-      const variant = parserOutput.components.first().getVariants().first();
+      const variant = parserOutput.components.first().getDefaultVariant();
       const view = parserOutput.components.first().getFiles().find({stem: 'view'});
       const opts = {
```
dkhuntrods: As above - we should mock the renderer here

packages/internals/renderer/src/adapter.test.js
```diff
@@ -28,16 +30,16 @@ describe('Adapter', function () {
...
         const adapter = new Adapter(Object.assign({}, validAdapter, {match}));
-        expect(adapter.match(funjucksFile)).to.equal(true);
-        expect(adapter.match(otherFile)).to.equal(false);
+        expect(adapter.match(funjucksFile.path)).to.equal(true);
+        expect(adapter.match(otherFile.path)).to.equal(false);
       });
     });
 
-    it('throws an error if the file argument is not a File instance', function () {
+    it('throws an error if the file argument is not a path', function () {
```
dkhuntrods: Does it expect an actual path-like-string here, or just a string? Might be worth renaming for clarity if latter.

packages/internals/support/src/validator.js
```diff
@@ -6,6 +6,11 @@ class Validator extends Ajv {
   constructor(...args) {
     super(...args);
     ajvKeywords(this, ['instanceof', 'typeof', 'regexp']);
+    const definition = require('ajv-keywords').get('instanceof').definition;
```
dkhuntrods: Is it worth adding the collections here?

allmarkedup: Yes definitely, will add that in as a future separate PR though.

packages/fractal/fractal/src/fractal.test.js
allmarkedup: Just added a commit that sorts this, stubs the methods rather than allowing a collections/renderers to be passed in as opts.

PR #357, opened 2017-09-21
Implement component.getView() and .getViews()

dkhuntrods:
TODO: Testing of integration with global config

### DIFF
--- a/packages/internals/support/src/entities/component.js
+++ b/packages/internals/support/src/entities/component.js
@@ -62,8 +62,7 @@ class Component extends Entity {
 
   getViews() {
     let views = new FileCollection();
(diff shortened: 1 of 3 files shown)


packages/internals/support/src/entities/component.js
```diff
@@ -61,11 +61,17 @@ class Component extends Entity {
   }
 
   getViews() {
-    // TODO
+    let views = new FileCollection();
+    const viewConfig = this.getConfig().views;
```
allmarkedup: just wondering.... should this get the `views` prop directly from the config or should it use `this.get('views.filter')`? I.e. if people override this or define a getter should that be the value that is used as the filter?

PR #358, opened 2017-09-22
Rename unclear properties and entities  

allmarkedup:
This PR renames a few things to make their intent clearer, differentiate from other similar items and make the system easier to document/explain. The main updates include:

- Changing component & variant `name` properties to `id` to better reflect the expectations around how it should be formatted and used
- Adapters are renamed to Engines, and the Renderer is simplified to be an EngineStore to remove code duplication and mixed responsibilities.


allmarkedup:
@dkhuntrods This PR is ready to go now for whenever you have a chance to take a look at it :-)

PR #365, opened 2017-09-27
Change order of <code><pre>... to <pre><code>... in markdown.js

siiron:
`<pre><code>...dacode...</code></pre>` is the correct order for code-blocks according to W3C
https://www.w3.org/TR/2011/WD-html5-author-20110809/the-code-element.html

### DIFF
--- a/src/core/markdown.js
+++ b/src/core/markdown.js
@@ -11,9 +11,9 @@ renderer.code = function (code, lang, escaped) {
         code = output;
     }
     if (!lang) {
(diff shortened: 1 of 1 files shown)


siiron:
By having `<code>`-tags outside `<pre>`-tags, you might get weird artifacts depending on how you style the single-line `<code>`. So by changing the order to [W3C's recommendations](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-pre-element.html#the-pre-element), these artifacts are gone:

order-of-pre-and-code

PR #384, opened 2017-10-24
Fractal plugin inspector assets

dkhuntrods:
Adds a Plugin to generate a (currently unsorted or grouped) FileCollection of assets on each component at `component.inspector.assets`


packages/fractal/fractal/src/config/defaults.js
```diff
@@ -19,6 +19,13 @@ module.exports = {
         views: {
           match: 'view.*',
           default: '.html'
+        },
+        assets: {
+          styles: '**/*.{css,scss}',
+          scripts: '**/*.{js}',
```
allmarkedup: Just testing this and it seems that the `scripts` glob here won't actually match any JS files - the `{}` syntax only works if there is more than one comma separated option within it, so this should actually be `**/*.js`

packages/fractal/fractal-plugin-inspector-assets/src/plugin.js
```diff
@@ -0,0 +1,98 @@
...
+
+    transform: 'components',
+
+    async handler(components, state, app) {
+      // FIXME: looping/compiling need to be optimised here
+      components = await components.mapAsync(async component => {
+        const outputFileSystem = new MemoryFS();
+        component.inspector = component.inspector || {};
+        const assets = component.getAssets();
+        if (!assets) {
```
allmarkedup: As `.getAssets()` returns a FileCollection I think this needs to check for `assets.length`? Otherwise will always return true.

Issue #461, opened 2018-08-27
Vulnerable dependency - growl@1.9.2 via mocha<@4.0.1


dkhuntrods:
PR #460 brought a critical growl vulnerability in the mocha package to our attention.

dkhuntrods:
Resolved by #462

PR #462, opened 2018-08-27
Update mocha and vorpal

dkhuntrods:
Addressing the security vulnerability in #461:
 
- Updates mocha to 4.0.1 to remove growl@1.9.2. 
See https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#401--2017-10-05.

- Update vorpal to 1.12.0 after npm audit found additional errors. 

Tested on 
- macOS 10.12.6

Node versions: 
- v6.9.4
- v8.11.4 
- v10.9.0


dkhuntrods:
Thanks!
Only incrementing to 4 purely because I'm trying to make the smallest
changes that are necessary. Once the security pass is done we can look at
that as part of updating all the packages.

PR #465, opened 2018-09-19
Update to nunjucks 3.1.3

khawkins98:
Adds a lot of useful features including formatting, for loops; so we can do things like:
`{% asyncEach item in variable %}`

Seems to not break anything, `npm test` passes, and I don't know of any breaking changes in Nunjucks 2 --> 3.


dkhuntrods:
Hi @khawkins98 thanks for this! 
Looks good - Would you mind updating your PR against master so I can get in merged in?

khawkins98:
👍 Pulled in latest from the origin, should be all set.

PR #468, opened 2018-09-20
Perform all minor/patch dependency updates.

Chapabu:
Updated Bluebird, Handlebars, and Lodash to their latest minor releases.

I only updated the smaller patch versions as there are around 24 dependencies that need major version updates. All the time there are no security releases for the other dependencies, I think we're better off leaving them until we've got better test coverage.

Also, if we're starting to think about V2 we may never even need to update the remaining packages.


dkhuntrods:
Thanks @Chapabu! Agreed with all points above, going to start working with @mikkelrom to get the integration test coverage improved but I'd say there's scope to improve the existing unit tests as well. I'll take this for a spin now.

Issue #464, opened 2018-09-07
Empty HTML panel


supermueller:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Checked the Fractal Documentation for information about your issue: https://fractal.build/guide
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Afrctl

### Description

If the notes panel is on first position and is not empty, then the HTML panel shows no content.

### Steps to Reproduce

1. Rearrange the panels in `fractal.js`

```js
const mandelbrot = require('@frctl/mandelbrot');

const myCustomisedTheme = mandelbrot({
    panels: [
        "notes", "html",
    ],
});

fractal.web.theme(myCustomisedTheme);
```
2. Start Fractal web UI server

3. Add a Note to a component by adding a file `readme.md` with some content.

4. The HTML code disappears in the HTML panel.

**UPDATE:** The HTML panel is only empty, if the component view includes another component by using the render helper, e.g.: 

```
<div class="this-component">
    {{render '@another-component' }}
</div>
``` 

**Expected behavior:**

The HTML panel should still show the HTML code.

**Actual behavior:**

The HTML panel doesn’t show the HTML code.

**Reproduces how often:**

100%

### Versions

* What version of Fractal are you running? (run `npm list @frctl/fractal` / `yarn list @frctl/fractal`

```bash
└─┬ @frctl/fractal@1.1.7 
  ├─┬ @frctl/handlebars@1.1.4
  │ └── @frctl/fractal@1.1.7  deduped
  └─┬ @frctl/mandelbrot@1.2.0
    └── @frctl/fractal@1.1.7  deduped
```

* What version of Node.js are you running? (run `node -v`)

```bash
v8.11.3
```

dkhuntrods:
Thanks @supermueller - I suspect this is a Mandelbrot (Fractal theme, rather than Fractal itself) issue, but we'll look into it.

pdreij:
Having the same problem, can be temporary solved by moving "notes" panel after "html".

LeBenLeBen:
Potentially related to https://github.com/frctl/nunjucks/issues/5

LeBenLeBen:
@supermueller On my side, the HTML panel is empty only if the component includes another one with the `render` helper. Can you confirm you have the same behavior?

[...]

PR #472, opened 2018-10-04
Circular reference fix

allmarkedup:
This PR should address #464 as well as frctl/nunjucks#5. The circular reference issue was caused by the markdown renderer directly mutating its supplied config object, which itself (as part of the complete fractal config object) is later made available to templates as a property of the `_env` variable.

A basic test has been included to confirm the md config object is now cloned before mutation. If anyone is still having issues after this fix then let me know :-)

### DIFF
--- a/src/core/markdown.js
+++ b/src/core/markdown.js
@@ -1,19 +1,22 @@
-'use strict';
+"use strict";
 
-const marked = require('marked');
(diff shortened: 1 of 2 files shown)


allmarkedup:
Sorry just realised this PR was based on the wrong branch - will close this one and fix

PR #479, opened 2018-11-21
Log Error when readFile fails to load a file

Schleuse:
There is currently no notification when `readFile` fails to load a file. Can easily happen when a file is moved, renamed and deleted.


Chapabu:
This looks great to me! Thanks @Schleuse :)

Issue #417, opened 2018-02-13
Fractal Crashes and doesn't recover automatically when using BrowserSync


jamesf-deeson:
**Setup:**
In our scenario we are using BrowserSync in tandem with a separate Webpack process, having Webpack compile our assets continuously and Fractal display them.  The command that we're using to kick this process off is ```fractal start --sync & webpack --watch```

**Problem:**
If we make a syntax error, which we do most commonly in our Fractal component config files (`component.config.js`) then Fractal will throw an error, eg:

```
✘ Error parsing data file /Users/james/repositories/components/03-components/promo/promo.config.js: SyntaxError: Unexpected token {
✘ SyntaxError: Unexpected token {
```

Unfortunately Fractal doesn't automatically recover from this when we do fix the syntax issue, which means stopping and restarting build process for the project, which is frustrating.  Webpack in contrast has no issues, and will happily detect that the files have changed and successfully recompile when the syntax is correct.

Is there some change that we can make to our build process to be more reliable, or some feature we're missing with Fractal which provides a better experience when using the `--sync` flag?

guschilds:
I've had the same issue on a couple of projects and for what it's worth, I think Webpack is unrelated.

**To reproduce:**

1. Start the Fractal server (whether with the `fractal` CLI or a Gulp command) with `sync` enabled.
2. Verify that valid changes to a `component.config.js` file is properly reflected when the page refreshes.
3. Introduce a JS syntax error into the `component.config.js` file and save it. From here, the console will log an error like the one in the issue description.

**Expected result:** After fixing the error and saving the file, Fractal continues to read from the file and if it is valid, any changes to the data are reflected in the browser.

**Actual result:** After fixing the error and saving the file, Fractal does not read from it again. The browser refreshes, but any valid changes are not reflected in the browser. To continue working, the Fractal server must be restarted before it will continue reading valid changes within the file. This can be frustrating and burn a lot of time.

---

A co-worker and I were able to trace the issue to a `Promise.reject()` call within `src/core/data.js`. As soon as that is called, changes to that file (valid or not) will no longer be read. Unfortunately I'm not familiar enough with the Fractal codebase, Node, Promises, etc. to find a solution that feels truly appropriate, but I was able to patch Fractal in a way that lets us know which file has an error without requiring the server to be updated. We're now using [`patch-package`](https://www.npmjs.com/package/patch-package) to create and install this patch:

```
patch-package
--- a/node_modules/@frctl/fractal/src/core/data.js
[...]

PR #509, opened 2019-08-08
Fixed sync crash when JS config files contain errors. (#417)

guschilds:
This solves the problem described in https://github.com/frctl/fractal/issues/417.

When using Fractal with JS-based config files (`*.config.js`), BrowserSync will crash and not recover if the config file has a syntax error in it. Even after fixing the syntax error, the browser will refresh, but still shows the last state of the component before the error occurred. It never reflects the new, valid state of the configuration. The only remedy is then to restart the Fractal server.

**Steps to reproduce:**

1. Create a Fractal project using the "TL;DR method" described [here](https://fractal.build/guide/getting-started.html#the-tl-dr-method).
2. Switch the `example.config.yml` file to `example.config.js`, while updating the syntax appropriately (you may need to restart the Fractal server after doing so):
```
module.exports = {
  title: "Example",
  context: {
    "text": "This is an example component!",
  }
};
```
3. Open `localhost:3000` in the browser, view the Example component, and verify that it displays as expected.
4. Make a valid change to `text`'s value within `example.config.js`, save the file, and verify that the change is automatically reflected in the browser.
5. Make an invalid change to `text`'s value within `example.config.js` (such as deleting the quotation mark at the end of the string) and save the file. Verify that an error is thrown in the command line and that the browser shows the last valid state of the component.
6. Fix the invalid change, such as adding the quotation mark back, and save the file. Also, make another change to `text`'s value.

**Expected result:** The browser should show the example component with the most recent change to `text`'s value.

**Actual result:** The browser shows `text`'s value from step 4 above and no amount of refreshing or fixing the file changes that. Only restarting the Fractal server and refreshing the page will.

**Actual result with this fix applied:** The browser will not show the component at all, but a helpful error is thrown in the command line. When the invalid change to the config file is fixed and the file is saved, the browser will refresh with the most recent changes to the file. The Fractal server never needs to be restarted.

An example of the error:

```
✘ Error parsing data file example.config.js: Invalid or unexpected token
```

This can save a huge chunk of time and confusion when developers are working with large, complex configuration files. More explanation behind how this solution was arrived upon is described in the issue linked above. The error message in this PR is slightly different than what was suggested there, but for the better.

Thank you for your work with Fractal!

### DIFF
--- a/src/core/data.js
+++ b/src/core/data.js
@@ -62,8 +62,8 @@ module.exports = {
                 }
                 return Promise.resolve(data);
             } catch (err) {
-                Log.error(`Error loading data file ${filePath}: ${err.message}`);
-                return Promise.reject(err);
+                Log.error(`Error parsing data file ${filePath.split('/')[filePath.split('/').length - 1]}: ${err.message}`);
+                return Promise.resolve({});
             }
         } else {
             return fs.readFileAsync(filePath, 'utf8').then(contents => this.parse(contents, format)).catch(err => {


Chapabu:
Thanks for this @guschilds :)

PR #513, opened 2019-09-04
fix: don't stop server when files are not found

Borange:
Notice if I have not set an favicon and running the CLI fractal start --sync. The fractal stop working in preview mode. This purposal will only stop the server if the error is not 404.


Chapabu:
This is almost the same fix as #515. Thanks for this! I merged the other PR instead as this one will cause the server to stop on errors which won't _actually_ kill Fractal.

PR #518, opened 2019-09-25
Export Log for usage in context

Schleuse:
It can be pretty useful to communicate warnings and such in context:

```javascript
const { log } = require("@fractal/fractal");
log.warn("Deprecated: this context-data should not be used anymore - Use ... instead.");
```

### DIFF
--- a/src/fractal.js
+++ b/src/fractal.js
@@ -127,6 +127,7 @@ module.exports.Fractal = Fractal;
 module.exports.WebTheme = require('./web/theme');
 module.exports.CliTheme = require('./cli/theme');
 module.exports.Adapter = require('./core/adapter');
+module.exports.log = require('./core/log');
 module.exports.utils = require('./core/utils');
 
 module.exports.core = {


Chapabu:
I'm not completely averse to this, I'm just curious as to what this offers over a component state instead. A log message won't be viewable _inside_ Fractal, whereas simply creating a new component state called "Deprecated" seems to offer more value.

Schleuse:
We're stacking styleguides (pattern and context) on top each other.

The base-styleguide offers more general components, whereas the customer styleguide uses or extends those base-components. 

Sometimes it is necessary to move or refactor those base components. This is a problem, when a customer project used these base-components.

In this case I'd like to offer a layer of compability:

 ```javascript
const { log } = require("@fractal/fractal");
log.warn("Deprecated: this context-data should not be used anymore - Use foobar instead.");`
module.exports = require("module/components/02-molecules/foobar/context.config.js");
```

Chapabu:
Sorry, this one escaped me! Thanks - all merged :)

PR #521, opened 2019-10-08
Upgrade to gray-matter 4

aredridel:
This preserves the toml and coffeescript handling. Some weight could be shed eliminating those features, but I went for compatibility instead.


aredridel:
I'd love to have this looked at!

PR #535, opened 2019-11-22
Catch error when parsing data for Component

Schleuse:
Fixes a issue where fractal won't start, when a component has an error in the data it uses.

Errors in component-data can easily happen when they are not written in `json` but in `js`


risker:
@Schleuse could you bring an example of a faulty config file and a fractal config that will cause fractal not to start? then we could test the solution with an actual test case.

either way, i think this try/catch should be implemented in the `Data.readFile` function, since it already tries to catch some of those errors.

Chapabu:
This was closed purely because the default branch was changed, and develop has now gone in favour of master (hence GitHub auto closes it).

Unfortunately, I can't re-open or repoint it, so this will have to be @Schleuse

PR #539, opened 2019-12-07
allow function values in context objects

risker:
I don't believe there's an issue for this feature, but I'll do my best to explain it here.

I'm using Fractal for multiple projects with React components in a way that the components are rendered on the server and hydrated on the client. React components can have props that can be assigned functions to. For example, HTML anchors and buttons can have onClick prop.

Now, given the following config file:
```
module.exports = {
    context: {
        text: "lorem ipsum",
        onClick: function(event) {
            console.log(event);
        },
    },
};
```

Currently, the context object returned by Fractal looks like this:
```
{
    text: "lorem ipsum",
    onClick: {}
}
```

So it basically just resolves the function to an empty object.

Following this change, functions in context objects are preserved, so the returned context object looks like this.
```
{
    text: "lorem ipsum",
    onClick: function(event) {
        console.log(event);
    }
}
```

### DIFF
--- a/src/core/resolver.js
+++ b/src/core/resolver.js
@@ -66,6 +66,10 @@ const resolver = module.exports = {
                 return null;
             }
 
+            if (_.isFunction(item)) {
+                return item;
+            }
+
             if (_.isArray(item) || _.isObject(item)) {
                 return resolve(item);
             }


b-connect:
This would be also awesome for Objects.
```js
if (_.isObject(item)) {
  return item;
}
```
This would call Object.prototype.toString method.

risker:
I've been thinking about this more, and I don't really believe this should be placed in core. I think it may be expected in some places that all config will be resolved to a JSON-compatible object, which this change would break.

Instead, maybe an option of some kind that allowed custom resolvers?

Issue #545, opened 2020-01-14
Server occasionally returns an empty response


AndrewLeedham:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Checked the Fractal Documentation for information about your issue: https://fractal.build/guide
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Afrctl

### Description

When loading a 'detail' page the preview iframe sometimes does not display a component and looking at the network panel in dev tools (FireFox) the response is empty:
![image](https://user-images.githubusercontent.com/5557458/72357006-26d19680-36e2-11ea-9cd9-f7f7b89af0b8.png)

### Steps to Reproduce
1. Create a fresh fractal project with Gulp https://fractal.build/guide/integration/build-tools.html#gulp and add a component.
2. Launch the server locally with `gulp fractal:start` and navigate to the detail page for the added component. (It should not show the component initially)
3. See empty response in the network panel in dev tools.

**Expected behavior:**
The preview iframe to load a component consistently.

**Actual behavior:**
Empty response on first page load.

**Reproduces how often:**
Nearly always on the first page load of the server session and sub-sequent navigations to other components until the first refresh.

### Versions

* What version of Fractal are you running?: **1.2.1**
* What version of Node.js are you running?: **12.9.1**
* What browser are you running?: **FireFox 70.0.1**
* What OS are you running?: **macOS Mojave 10.14.6**

### Additional Information

Having experimented locally I believe the issue is due to a race condition where `end` is being called before the `send` request is fulfilled (https://github.com/frctl/fractal/blob/develop/src/web/server.js#L240). Looking at the express documentation `send` handles the full request so `.end()` is not necessary. Removing `.end()` fixes the issue for me locally.

AndrewLeedham:
Turns out the `0B` I was seeing in dev tools was a very hard to debug iframe quirk which did not log anything, it was caused by something in my custom theme. Therefore, I will close the issue for now. However, the PR is still relevant given the Express docs, also we run VRT against the fractal server which has some occasional failures, so I am going to experiment and see if the Express change resolves that.

PR #546, opened 2020-01-14
Remove end from server rendering

AndrewLeedham:
Removes `.end()` since `.send()` already handles this:

> "Use to quickly end the response without any data. If you need to respond with data, instead use methods such as res.send() and res.json()." - https://expressjs.com/en/4x/api.html#res.end 

Will hopefully address some occasional failed responses. See #545 for context.


risker:
Thanks and sorry for the mess here - tried to update your fork but messed something up in the beginning i think.

Issue #557, opened 2020-03-23
1.2.1 Markdown Code Fail


dutscher:
Hi There,

we are currently on 1.1.7 and saw that there was an release 1.2.1
what i found is a bug in the markdown.js file:

https://github.com/frctl/fractal/compare/v1.1.7...v1.2.1#diff-6babe5ef701ed6699152e572b17e31c9

in 1.1.7 it was:
```<code><pre>${code}</pre></code>```
in 1.2.1 it is:
```<pre><code>${code}</code></pre>```

and this will do simple markdown code ` ```<html ...>``` ` to single line code.

1.1.7 markdown
![image](https://user-images.githubusercontent.com/14682/77325543-cd589a00-6d18-11ea-8019-79c734d2c3ee.png)
looks now with 1.2.1 same correct markdown:
![image](https://user-images.githubusercontent.com/14682/77325625-eb25ff00-6d18-11ea-9a1f-157c853b1e17.png)

cheers

risker:
This is a duplicate of #532, which was fixed by frctl/mandelbrot#79

PR #558, opened 2020-03-23
revert code markdown tag order

dutscher:
see issue https://github.com/frctl/fractal/issues/557


risker:
will not accept this, as this is resolved by a style update in mandelbrot, see frctl/mandelbrot#79

PR #563, opened 2020-03-31
add exclude config option to components

### LINKED ISSUE #556: fractal includes node_modules on build and shows it on styleguide
### Description and steps to reporduce

I have made an example repo of current issue on: https://github.com/laurikihva/fractal-demo (setup is described in readme.md). What I basically do there is that I make a component as separate module on `project-1` and then use this component as module on `project-2`. If you run project-2 locally with `npm run dev:project-2` then under components there are also `node_modules` which actually shouldn't be shown there by fractal. Also if you run project-2 then in `/components/detail/tabs` example it renders context as "my awesome tabs item from project 1" but it should actually render "my awesome tabs item from project 2" which is set in @project-2/tabs.config.yml file, if I change the name of variant, then it works as intended.

**Expected behavior:** Fractal should ignore `node_modules` folder on building styleguide which means not rendering it and not setting `viewPath`'s from it.

**Actual behavior:** Fractal renders `node_modules` in `project-2` and also gets `node_modules` `viewPath` from `project-1` which doesn't get proper context from project-2 config files in styleguide.

**Possible solution:** One possible solution for us could be that we can pass a setting to fractal which files or folders it should ignore on building, for example it should ignore `node_modules` folder.

### Versions

* What version of Fractal are you running? @frctl/fractal@1.2.1
* What version of Node.js are you running? v10.15.0

risker:
This PR will resolve #556.

Following this change, we will allow users to exclude some directories from Fractal structure like so:
```js
fractal.components.set('exclude', '**/node_modules/**');
```

### DIFF
--- a/src/api/components/source.js
+++ b/src/api/components/source.js
@@ -341,12 +341,18 @@ module.exports = class ComponentSource extends EntitySource {
     }
 
     isView(file) {
-        return anymatch([
+        const matchers = [
             `**/*${this.get('ext')}`,
             `!**/*${this.get('splitter')}*${this.get('ext')}`,
             `!**/*.${this.get('files.config')}.${this.get('ext')}`,
             `!**/${this.get('files.config')}.{js,json,yaml,yml}`
-        ], this._getPath(file));
+        ];
+
+        if (this.get('exclude')) {
+            matchers.push(`!${this.get('exclude')}`);
+        }
+
+        return anymatch(matchers, this._getPath(file));
     }
 
     isVarView(file) {


src/api/components/source.js
```diff
@@ -341,12 +341,18 @@ module.exports = class ComponentSource extends EntitySource {
...
-        return anymatch([
+        const matchers = [
             `**/*${this.get('ext')}`,
             `!**/*${this.get('splitter')}*${this.get('ext')}`,
             `!**/*.${this.get('files.config')}.${this.get('ext')}`,
             `!**/${this.get('files.config')}.{js,json,yaml,yml}`
-        ], this._getPath(file));
+        ];
+
+        if (this.get('exclude')) {
```
Chapabu, line 351: I think we need a type check here. Someone could add anything in the config here.

Can we also open an issue to update the docs before this gets merged?

Chapabu:
This PR was fine, but needs resubmitting against `master` as the default project branch has changed and develop doesn't exist any more.

PR #564, opened 2020-04-01
chore: update anymatch, chokidar and nunjucks

levito:
Update `anymatch`, `chokidar` and `nunjucks` to remove `braces@1.8.5` which has a ReDoS vulnerability (fixed in v2.3.1)


levito:
Hi @risker, I downgraded `anymatch` to version 2.0.0. This is still based on micromatch so it should not change the way files are matched. But it already fixes the security issue.

Can you please check if this fixes the problem on Windows?

risker:
Thanks, looking good now.

PR #566, opened 2020-04-03
Teach variants to read notes from component--variant.notes.md

rchavik:


risker:
The implementation for this should go through the file matchers in ComponentSource, similarly to how variant views are handled - then passed to VariantCollection, that should pass the correct File object to Variant. The files are parsed into Fractal through ComponentSource, so it should not be necessary to use `fs` directly in here. Currently the additional md files would also be registered as component resources, though they shouldn't - the default readme.md file does not.

Additionally, there should be a discussion on how the file name should be. Currently, we allow compound components to have a `[files.notes].md` file, where `[files.notes]` is configurable through config and defaults to `readme`.

Considering this, the following possibilities are on the table:
1. the variant readme file should be named `[files.notes]--[variant.name].md`. For example, `readme--secondary.md`.
2. the variant readme file should be named `[component.name]--[variant.name].[files.notes].md`. For example, `button--secondary.readme.md`. If we go this route, we should also allow `[component.name].[files.notes].md` (ex `button.readme.md`) as the default readme file for consistency.

I personally lean towards 2, but 1 should be fine as well.

rchavik:
I like option 2 as well. Will try to implement it.

rchavik:
Updated the PR for option 2.

src/api/components/source.js
```diff
@@ -435,6 +435,8 @@ module.exports = class ComponentSource extends EntitySource {
 
             if (view) { // it is a component
                 const nameMatch = dir.name;
+                const notes = matchFile(file => anymatch(`**/${nameMatch}.${source.get('files.notes')}.md`, source._getPath(file)) );
```
risker, line 438: this should use the existing `isReadme` matcher - they're all abstract in the sense that they do not need to match a specific component.

src/api/components/source.js
```diff
@@ -435,6 +435,8 @@ module.exports = class ComponentSource extends EntitySource {
...
+                dirConfig.notes = notes && notes.isFile ? notes.getContentSync() : undefined;
```
risker, line 439: - should not assign `dirConfig.notes` here - the readme should already be discovered by the matcher when `dirDefaults` is assigned
- the isFile/getContentSync is redundant since in Component `get notes()` already does that.

src/api/variants/collection.js
```diff
@@ -139,6 +139,7 @@ module.exports = class VariantCollection extends EntityCollection {
 
         views.filter(f => !_.includes(usedViews, f.base)).forEach(viewFile => {
             const name = utils.slugify(viewFile.name.split(source.get('splitter'))[1]).toLowerCase();
+            const notes = resources.find(r => r.base === `${component.name}${source.get('splitter')}${name}.${source.get('files.notes')}.md`);
```
risker, line 142: this should be passed in via a separate parameter (like `views` is), since the variant readme files should not be considered resources at all.

src/api/variants/collection.js
```diff
@@ -147,6 +148,7 @@ module.exports = class VariantCollection extends EntityCollection {
                 order: viewFile.order,
                 dir: opts.dir,
                 isHidden: viewFile.isHidden,
+                notes: notes && notes.isFile ? notes.getContentSync() : undefined,
```
risker, line 151: - pass the readme file as `readme` key, since at this stage Fractal considers notes as coming from config file & readme as separate file
- the isFile/getContentSync would be better inside Variant class `get notes()` getter, so the implementation would be more similar to Component class

src/api/components/source.js
rchavik: Currently, `isReadme` only match against `readme.md` and has no knowledge of the current component (cmiiw). How do we support notes as option 2 described, ie: `[component.name].[files.notes].md`?

risker: the matcher function is abstract, it matches files via globs.
so if currently we have `**/${this.get('files.notes')}.md` as the matcher in isReadme, we should add a second matcher `**/*.${this.get('files.notes')}.md`. Note the extra `*.`, which should now match the new file name structure.

src/api/variants/collection.js
```diff
@@ -129,6 +130,8 @@ module.exports = class VariantCollection extends EntityCollection {
             p.viewPath = Path.join(p.dir, p.view);
             p.handle = `${component.handle}${source.get('splitter')}${p.name}`.toLowerCase();
             p.isHidden = _.isUndefined(conf.hidden) ? viewFile.isHidden : conf.hidden;
+            const readmeName = `${opts.viewName}${source.get('splitter')}${p.name}.${source.get('files.notes')}`.toLowerCase();
+            p.readme = _.find(readmes, f => { let ret = f.name.toLowerCase() === readmeName; if (ret) {console.log({f})}; return ret;});
```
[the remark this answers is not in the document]
rchavik: Yeah..  sorry about that

rchavik:
Thanks for the pointers.

I have tried to adjust and tested your manual test cases. It seems to work now.

Not really sure about the non compound component support, I'll leave a note in the corresponding file.

src/api/components/source.js
```diff
@@ -471,12 +483,14 @@ module.exports = class ComponentSource extends EntitySource {
                     viewName: view.name,
                     viewPath: view.path,
                     dir: dir.path,
+                    readme: readmeFile ? new File(readmeFile) : undefined,
```
rchavik, line 486: i need to pass a `new File()` instance, otherwise there's `getContentSync()` is not a function error.

risker: yeah, we do need a File instance. could we reuse the `matchFile` function here? so `readme: matchFile(source.isReadme)` could maybe do the trick?

src/api/variants/collection.js
```diff
@@ -129,6 +132,8 @@ module.exports = class VariantCollection extends EntityCollection {
             p.viewPath = Path.join(p.dir, p.view);
             p.handle = `${component.handle}${source.get('splitter')}${p.name}`.toLowerCase();
             p.isHidden = _.isUndefined(conf.hidden) ? viewFile.isHidden : conf.hidden;
+            const readmeName = `${opts.viewName}${source.get('splitter')}${p.name}.${source.get('files.notes')}`.toLowerCase();
```
risker, line 135: this readmeName assignment + find filter on the next line is duplicated in three places. let's turn this into a helper function and reuse it.

Chapabu:
I'm happy with this.

I've opened https://github.com/frctl/fractal-docs/issues/38 as we'll need to ensure we update the documentation.

PR #568, opened 2020-04-06
add coverage and unit tests for collection mixin class

risker:
I'm trying to up my test writing skill and this project needs some tests, so here's a first draft.

Coverage-wise the situation is a bit better than it was before, although I'm not really sure how relevant all those tests are. I guess there could be more expects here and there, suggestions are welcome.

Skipped squash method in that mixin class, since I couldn't really figure out the expectations for that - as far as I looked, it isn't documented nor used anywhere. Wondering if we could mark it as deprecated and/or remove immediately...


Chapabu:
Looks good. In general, I think we're testing a bit too much here, but that may just be because we have too much code for very simple things in general.

Issue #438, opened 2018-05-11
Inject sub-components from context 


pdureau:
Hi, 

Thanks so much for your work on Fractal.

I am the maintainer of the Drupal integration of Fractal: https://www.drupal.org/project/ui_patterns_fractal

According to the Fractal documentation, we need to set the sub-components inside the template file: 
```
<div class="searchbox">
    <input type="search" name="keywords">
    {{> @button }}
</div>
```
https://fractal.build/guide/components/sub-components

It would be very useful to be able to keep those templates "sub-components agnostic", by specifying a simple variable:
```
<div class="searchbox">
    <input type="search" name="keywords">
    {{ submit }}
</div>
```
And inject the sub-component from the description file context. For example:
```
title: Search Box
context:
  submit: @button
```
Other proposal, with context overriding:
```
title: Search Box
context:
  submit: 
    render: @button
    text: Go!
```

For Drupal integration, it is a must-have, because its render system works like that: https://www.drupal.org/docs/8/api/render-api/render-arrays

And I strongly believe this mechanism would be useful even for non-Drupal people.

Is it already possible with Fractal?

PS: I am using the Twig engine by the way: https://github.com/frctl/twig

pdureau:
Such a mechanism is available in Pattern Lab + Twig, with Data Transform Plugin.

#### Include pattern files

If a value contains the name of a pattern in shorthand partials syntax, the plugin will replace the value with the rendered pattern:

```json
{
    "key": "atoms-form-element-label.html"
}
```

Advanced syntax with support for passing variables and disabling access to the default data:

```json
{
    "key": {
        "include()": {
            "pattern": "atoms-form-element-label.html",
            "with": {
                "title": "Textfield label"
            },
            "only": true
        }
    }
}
```

In both examples the value of `key` will replaced with the rendered pattern.


#### Include pseudo-pattern files

It is also possible to include [pseudo-patterns](http://patternlab.io/docs/pattern-pseudo-patterns.html) using the shorthand partials syntax, by replacing the tilde (~) with a dash (-). So for example the pseudo-pattern `shila-card.html~variant.json` can be included like so:

```json
{
    "key": "molecules-shila-card-html-variant"
}
```


#### Join text values

```json
{
    "key": {
        "join()": [
            "molecules-comment-html",
            "<div class=\"indented\">",
            "molecules-comment-html",
            "</div>",
            "molecules-comment-html"
        ]
    }
}
```

[...]

PR #573, opened 2020-04-13
Allow sub-components to be rendered fully

justafish:
Currently you can only reference and render context items contained within sub-components (e.g. `@menulocaltask.link`). Referencing `@menulocaltask` causes `[object Object]` to be rendered. This pull request adds the ability to reference `@@menulocaltask`, which will enable it to render the full component when no sub-context is specified with dot notation.

This is useful for template engines/styleguides which can't use partials to include other components. For example, Drupal's theming system can have other partials represented as regular twig variables.

Example:

```
label: "Menu Local Tasks"

context:
  primaryTabsTitle: "Primary Tabs"
  primary:
    - "@@menulocaltask"
    - "@@menulocaltask--active"
```

### DIFF
--- a/src/core/resolver.js
+++ b/src/core/resolver.js
@@ -40,6 +40,15 @@ const resolver = module.exports = {
                 return item.replace(/^\\@/, '@');
             }
 
+	        if (_.isString(item) && _.startsWith(item, '@@')) {
+                const entity = source.find(item.substring(1));
+                const fullRenderedComponent = source.engine().render(entity.viewPath, entity.content, entity.context, {
+                    self: entity.toJSON(),
+                    env: {},
+                });
+                return fullRenderedComponent;
+            }
+
             if (_.isString(item) && _.startsWith(item, '@')) {
                 const parts = item.split('.');
                 const handle = parts.shift();


risker:
i think the general idea here is fine, although i'm not sure if we want to add the extra syntax to Fractal core. maybe the better, more extendable & futureproof way would be to allow customizing the resolver?

the standard way to achieve logic like this would be to put objects into the array, that have both the component name and the data, for example:
```yaml
context:
  primary:
   - component: 'menulocaltask'
     data: '@menulocaltask'
```
then, in a template, you could do:
```twig
{% for item in primary %}
    {% include ('@' ~ item.component) with item.data %}
{% endfor %}
```
i guess i'm curious as to why that is not possible, why do some engines not allow using partials?

not rejecting this btw, just trying to understand the needs & find out if maybe there's a better way to accomplish this.

justafish:
Any template system which doesn't render top down will likely have issues, but to take a specific example from Drupal:
Menu Lock Tasks: https://git.drupalcode.org/project/drupal/-/blob/8.8.x/core/modules/system/templates/menu-local-tasks.html.twig
An individual task: https://git.drupalcode.org/project/drupal/-/blob/8.8.x/core/modules/system/templates/menu-local-task.html.twig

In menu-local-tasks you'll see that the menu item itself is represented by `{{ primary }}` and `{{ secondary }}`

risker:
Thanks for the example! This looks very similar to the issues I'm having in a Fractal+React project where we would need to add subcomponents in the `children` context property, but can't really since that is quite complicated to render/hydrate properly. I'll try to test this in my React usecase, if maybe it'll solve that one as well.

Chapabu:
@risker im happy to change the syntax slightly at this point. Rather selfishly because I use Drupal on a daily basis and can see how this may be useful to people.

However, customisation of the resolver is probably a better solution, and something we can consider when we star properly thinking about our big rewrite.

I'm on my tablet at the moment, but would someone be able to add an issue to the docs repo to slide this one in somewhere?

risker:
I filed frctl/fractal-docs#43

roryashfordbentley:
This looks like an promising solution to #438, I appreciate that there is an open ticket to write documentation but I was wondering how I can pass context for each sub-component from the component config. 

At the moment I have the following config.json for a component:

```json
{
    "context": {
        "title": "Locations A-Z",
        "description": "Filter the content by the first letter of the place name e.g. enter L for Leeds",
        "letters": [
            "@@a-z-button",
            "@@a-z-button"
        ]
    }
}
```

For each of the buttons I include is it possible to specify the context? So that in this instance I can pass a letter value through to the component e.g.

```json
{
    "context": {
        "title": "Locations A-Z",
        "description": "Filter the content by the first letter of the place name e.g. enter L for Leeds",
        "letters": [
            {
                render: "@@a-z-button",
                context: {
                    "letter": "A",
                    "link": "#a"
                }
            }
        ]
    }
}
```

LeBenLeBen:
@roryashfordbentley Yes, this is possible, there's an example in [this comment](https://github.com/frctl/fractal/pull/573#issuecomment-613050409) above using Yaml for config.

I don't know which templating language you’re using but with Nunjucks your context could be rendered like:

```jinja
{% for letter in letters %}
  {% render letter.render, letter.context, true %}
{% endfor %}
```

The third parameter, `true`, means it will merge with the component context.

PR #574, opened 2020-04-16
resolve date in config to JSON string

### LINKED ISSUE #455: YAML's date type is incorrectly parsed by Fractal
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Checked the Fractal Documentation for information about your issue: https://fractal.build/guide
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Afrctl

### Description

Sorry if this is a duplicate or I've missed something obvious, my Google-fu has failed me thus far.

When given a piece of context data containing an ISO formatted date, Fractal seems to return an empty object instead of the context data. I'm assuming this is because ISO 8601 format timestamps are considered a basic type in YAML and Fractal doesn't know how to parse them into a JavaScript equivalent. 

### Steps to Reproduce

Create a component with YAML-formatted context data including an ISO timestamp. For example:

```
context:
  updated: 2018-08-15
```

In the component, try rendering the context data with `{{ updated }}`. For me this is returning the string `[object Object]`.

Dumping the data with `{{ updated | dump }}` returns an empty object: `{}`. 

I've had the same result with other ISO timestamps such as `2018-08-15 17:13:00` and `2018-08-15 17:13:00+01:00`. Currently I can bypass this behaviour by quoting the timestamp, turning it into a string. 

Ideally the context data would be returned as a string, if not an object with date information. I'm also using the [Nunjucks adaptor](https://github.com/frctl/nunjucks), though I wouldn't expect that to be changing how the data is interpreted, just how it's templated (you know what they say about assumptions, though!) 

### Versions

Fractal version: 1.1.7
Node version: 10.4.0

risker:
Resolves #455 

Now, when using:
```yaml
context:
  date: 2020-01-01
```
it will be resolved to `2020-01-01T00:00:00.000Z`.

We could also resolve it to string (`Wed Jan 01 2020 02:00:00 GMT+0200 (GMT+02:00)`), but I think ISO timestamp is more correct in this case. I believe everything else in context is resolved to JSON as well in the end.

### DIFF
--- a/src/core/resolver.js
+++ b/src/core/resolver.js
@@ -66,6 +66,10 @@ const resolver = module.exports = {
                 return null;
             }
 
+            if (_.isDate(item)) {
+                return item.toJSON();
+            }
+
             if (_.isArray(item) || _.isObject(item)) {
                 return resolve(item);
             }


Chapabu:
I'd much rather just drop support for YML to be honest 😜

My only real worry with this is it's really a breaking change as it behaved like this for the duration of the project, so we'd need a new major release. Then it begs the question... What do we do with the other packages? We'd also need to check the state of npm, as we've got some hanging releases that we can't get rid of, so our next major release may actually need to be v4.

Leaving this for now, but we should discuss in Discord when we get a chance.

risker:
@Chapabu can you explain what exactly about this change is breaking? I don't see how the current behavior of returning a string `"[object Object]"` could be useful, or expected result for anyone. It also does not break the workaround (e.g. wrapping the date with quotes to make it a string for yaml), so at the very worst this changes the bug description, but at the best this resolves the issue.

LeBenLeBen:
Just had to jump it to share my love for yaml 🥰

Apart from that

Chapabu:
@risker Aah, what's breaking is my ability to read the linked issue. I see what you mean, this is fine!

Issue #576, opened 2020-04-17
Warnings in console caused by missing Highlight.js language


LeBenLeBen:
Using the latest version, I can see the following warning popping in the console when the server starts and when it reloads after file changes:

> (node:85075) Warning: a promise was created in a handler but was not returned from it, see http://goo.gl/rRqMUw
> [0] Could not find the language 'text', did you forget to load/include a language module?

I’ve see this issue in two different projects.

I’ll try to investigate to see if I can find the origin of it.

risker:
What's the version of installed highlight.js in your lockfile? I _think_ you somehow got highlight.js v10 beta installed, which changes how "text" is parsed:
> nohighlight or no-highlight are the only two CSS classes that will SKIP highlighting completely. *text* and *plain* no longer will do this.

maximedaoust:
@risker I have currently the same warning 

`Could not find the language 'vue', did you forget to load/include a language module?`

the current version in the lock file is the following released this on Feb 1
``` 
    "highlight.js": {
      "version": "9.18.1",
      "resolved": "https://registry.npmjs.org/highlight.js/-/highlight.js-9.18.1.tgz",
      "integrity": "sha512-OrVKYz70LHsnCgmbXctv/bfuvntIKDz177h0Co37DQ5jamGZLVmoCVMtjMtNZY3X9DrCcKfklHPNeA0uPZhSJg=="
    },
```

LeBenLeBen:
@risker I've just checked and have 9.18.1 in both main repo and theme.

LeBenLeBen:
@maximedaoust According to [this list](https://github.com/highlightjs/highlight.js/tree/master/src/languages) Highlight.js does not support `vue` as a language. You probably need an extension like https://github.com/highlightjs/highlightjs-vue

maximedaoust:
> @maximedaoust According to [this list](https://github.com/highlightjs/highlight.js/tree/master/src/languages) Highlight.js does not support `vue` out of the box. You probably need an extension like https://github.com/highlightjs/highlightjs-vue

thanks, I'll look into it.

risker:
It's a bit weird to debug this since it's hard to reproduce without code :)

Are you by any chance using code blocks in docs files? E.g.
```
\```text
codeee
 \```
```
That should be incorrect, and should use `plaintext` instead.

[...]

PR #579, opened 2020-04-24
Fix nunjucks view syntax

levito:
This makes view panels render templates with the following file extensions in Django mode (official replacement mode for Nunjucks): .nunjucks, .nunjs,. .nunj, .njk, .nj, .jinja, .jinja2, .j2

It removes the `langMap` from `src/core/highlighter.js` which worked inconsistently and moves the mapping logic to `src/core/utils.js`. This makes it work consistently and prevents warnings when using Nunjucks and running fractal or doing a static export.

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -17,6 +17,7 @@ module.exports = {
             case 'NUNJUCKS':
             case 'NUNJS':
             case 'NUNJ':
+            case 'NJK':
             case 'NJ':
             case 'JINJA2':
             case 'J2':
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -30,7 +30,7 @@ module.exports = {
             default:
                 return fang(filePath) || {
                     name: name,
-                    mode: 'text',
+                    mode: 'plaintext',
                     scope: null,
                     color: null,
                 };
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -12,12 +12,28 @@ const Stream = require('stream');
 module.exports = {
 
     lang(filePath) {
-        return fang(filePath) || {
-            name: Path.parse(filePath).ext.replace('.', '').toUpperCase(),
-            mode: 'text',
-            scope: null,
-            color: null,
-        };
+        const name = Path.parse(filePath).ext.replace('.', '').toUpperCase();
+        switch (name) {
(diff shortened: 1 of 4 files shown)


src/core/utils.js
```diff
@@ -12,12 +12,27 @@ const Stream = require('stream');
...
+                return {
+                    name: 'HTML+Django',
+                    mode: 'django',
+                    scope: 'text.html.django',
+                    color: null,
+                };
+            default:
+                return fang(filePath) || {
+                    name: name,
+                    mode: 'text',
```
LeBenLeBen, line 31: According to the issue we're having [here](https://github.com/frctl/fractal/issues/576), Highlight.js now requires this to be `plaintext`.

src/core/utils.js
```diff
@@ -12,12 +12,27 @@ const Stream = require('stream');
...
-            scope: null,
-            color: null,
-        };
+        const name = Path.parse(filePath).ext.replace('.', '').toUpperCase();
+        switch (name) {
+            case 'NUNJUCKS':
+            case 'NUNJS':
+            case 'NUNJ':
+            case 'NJ':
+            case 'J2':
```
LeBenLeBen, line 21: I guess this is missing the `njk` extension, [recommended by the Nunjucks docs](https://mozilla.github.io/nunjucks/templating.html).

LeBenLeBen: Or is it already supported by fang?

levito: @LeBenLeBen yes, it's in fang 2.0

src/core/utils.js
levito: Good point, I'll have a look at this tomorrow.

src/core/utils.js
levito: I had to revert to fang 1.0 since fang 2.0 returns a different object (no `mode` field but `ace_mode` and `codemirror_mode`. I don't want to this PR to be bigger than it needs to be, so including `NJK` is the better option.

risker:
LGTM, but let's wait for @LeBenLeBen to take a look at it again :)

levito:
@LeBenLeBen could you please take a look and merge if this is fine for you?

LeBenLeBen:
Thank @levito 🙇

PR #581, opened 2020-04-27
allow and prefer fractal.config.js as configuration file name

### LINKED ISSUE #580: Allow the config file to be called fractal.config.js
Currently, the default fractal config file is called `fractal.js`. There are numerous issues with calling the config that - most importantly, we even point it out in our own README that this is flaky on Windows and point to a solution that would point fractal to look at `fractal.config.js` instead.

Currently, users can opt out of the default config file name by specifying the fractal main file in their package.json like so:
```json
{
  "fractal": {
    "main": "fractal.config.js"
  }
}
```

I propose the following:
- Allow the config file to be named `fractal.config.js` (without the package.json workaround).
- Deprecate the `fractal.js` config file name, so we could remove support for it in the next major version. This should include a warning logged to the user's console informing them about the deprecation.

risker:
Resolves #580.

### DIFF
--- a/bin/fractal
+++ b/bin/fractal
@@ -90,7 +90,7 @@ FractalCli.launch(config, function(env){
         }
 
         if (Path.basename(configPath) === 'fractal.js') {
-            console.log(`Deprecated configuration file name! Rename your config file to fractal.config.js.`);
+            console.log(`Deprecated configuration file name! Rename fractal.js to fractal.config.js.`);
         }
 
     } else {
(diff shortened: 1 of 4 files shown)


bin/fractal
```diff
@@ -88,6 +89,10 @@ FractalCli.launch(config, function(env){
             });
         }
 
+        if (Path.basename(configPath) === 'fractal.js') {
+            console.log(`Deprecated configuration file name! Rename your config file to fractal.config.js.`);
```
LeBenLeBen, line 93:
```suggestion
            console.log(`Deprecated configuration file name! Rename fractal.js to fractal.config.js.`);
```

risker: good call :)

risker:
Yup, filed frctl/fractal-docs#42

PR #584, opened 2020-04-29
update handlebars adapter

risker:
the pretest command in symlink-dir basically self-requires the package to make the tests succeed. without it, in the context of this repository, mandelbrot can't require `@frctl/fractal` since it does not exist in node_modules anymore. not sure if this is the best or even correct solution, but it seems to work for now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -322,1119 +322,11 @@
         "to-fast-properties": "^2.0.0"
       }
     },
-    "@frctl/fractal": {
-      "version": "1.1.7",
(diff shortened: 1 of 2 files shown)


Chapabu:
Hmm.. That does feel a bit gross. I guess if it works that's fine, we can always revisit it later. The tests are practically useless at this point anyway unfortunately.

schnubb:
Hi,
I probably found a problem with the adapter version 1.2.0.
There was a change, where arrays should be ignored on merge, but this change overrides the default context for a rendered component.
Anyone seen this?

LeBenLeBen:
@schnubb could you please open a bug report with a detailed example?

PR #587, opened 2020-05-06
Move to multiple issue templates

LeBenLeBen:
This is a proposition to:
- Have an issue template for bugs, feature requests, and an extra link to the Discord for questions and support
- Globally hide guides for the reporter within comments so submitted issues are cleaner


Chapabu:
Funnily enough I was thinking something similar to this the other day.

I love the idea, but I'm also wary that the current issue template is way too verbose, and may put people off submitting issues. Might be worth chatting in Discord (or here), but I really think it's worth stripping them down

I think this bit should be moved to the readme under a "submitting issues" or "Got an issue?" heading at the very least:

> Have you read Fractal's Code of Conduct (https://github.com/frctl/fractal/blob/master/.github/CODE_OF_CONDUCT.md)? By filing an issue, you are expected to comply with it, including treating everyone with respect.
Do you want to ask a question? Are you looking for support? The Fractal Discord server is the best place for getting support:
https://discord.gg/vuRz4Yx

risker:
I'm fine with this, although I agree that complex issue templates always make submitting issues sooooo complicated and I personally never know what I should fill and what parts are optional. For example, "steps to reproduce" is mostly redundant if "expected behavior" and "actual behavior" are filled properly.

LeBenLeBen:
@risker @Chapabu I've updated it with simpler templates and an extra block in the README. WDYT?

PR #589, opened 2020-05-06
chore: update istextorbinary to get newer extensions

### LINKED ISSUE #588: Some text file formats are not displayed within the Assets panel
### Expected behavior

Files using popular and recent file extensions such as .vue should be displayed in the "Assets" panel.

### Actual behavior

Fractal says it cannot parse/display these files.

1. Add `resources` to the `panels` settings of Mandelbrot in your Fractal install
2. Add a .vue file in a component folder
3. Go to this component page, open the "Assets" panel
4. Select the .vue file in the dropdown

### Context

-   Fractal version: 
-   Node version: 12.14
-   OS: macOS 10.15.4

LeBenLeBen:
Fixes #588

It's a major bump but according to [their changelog](https://github.com/bevry/istextorbinary/blob/master/HISTORY.md#v300-2019-november-18), it's because of the drop of Node versions below 8, which we already doesn't support anymore.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1480,9 +1480,9 @@
       "integrity": "sha1-RqoXUftqL5PuXmibsQh9SxTGwgU="
     },
     "binaryextensions": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/binaryextensions/-/binaryextensions-2.1.1.tgz",
-      "integrity": "sha1-MgmlHKSkrVQaO409am1bg6JIWTU="
+      "version": "2.2.0",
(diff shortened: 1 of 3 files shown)


risker:
This is generally fine, although we also have a TODO in code related to this: https://github.com/frctl/fractal/blob/master/src/core/fs.js#L13

Ideally I'd remove that completely, I don't get why we should allow extensions that are not in that public repo. Or, at least, remove extensions from there that are supported by that repo.

`istextorbinary` also supports promises via `isBinaryPromise` since 2.5.0 - i think we should upgrade to that instead of doing our own promisify.

PR #592, opened 2020-05-12
chore: convert repository to monorepo

risker:
This is pretty much it :) I would like to import other repositories via separate pull requests, so this one only creates the necessary structure and moves the current core package to a subdirectory. As an alternative, we could keep this branch as our temporary working branch and I could point other import PRs here - then we could drop it all onto master with one merge - WDYT?

Some notes:
- I thought about adding a separate `examples/*` folder for some simple project examples - this could also be something we could test against. But I thought that that may as well be another PR.
- I opted for complete hoisting in root. In my experience this is better because this way there is less confusion about where to install dependencies. Of course, we can't completely enforce it by lerna, but that would be a hard guideline. I've also added a few words to the README about this.
- I configured the release script to be a drop-in replacement for `npm version` but use conventional commits. I really like conventional commits as a standard and I think it will make managing versions a charm. Using the option means that `npm run release` will choose the appropriate next version automatically based on what has been merged since last release. It also creates/updates the changelog for each package and can be configured to automatically create releases here on GitHub.
- Currently I've opted for independent versioning since our package versions are all over the place - but I think it's worth discussing if we want to switch to fixed versions at some point in the future.

### To Do
- [x] convert core repo to monorepo structure
- [x] import @frctl/mandelbrot
- [x] import @frctl/handlebars
- [x] import @frctl/twig
- [x] import @frctl/nunjucks

we also have the following packages that i'm not sure we should import at this point:
- @frctl/mustache (not updated since 2016)
- @frctl/react-adapter (wip - i think wip packages are ok when they are developed in separate repos - wdyt?)
- @frctl/consolidate
- @frctl/vash

### Merge strategy
1. This PR should be squash merged to master
2. Other PRs should be rebase merged to master so the commit history would be preserved. In the following order:
2.1. mandelbrot first, since that package has the most dependencies. The adapter PRs are pretty easy to redo from scratch.
2.2. nunjucks
2.3. twig
2.4. handlebars - this will have conflicts.

### Testing
- [x] I've merged all the monorepo branches together into `example/monorepo`. Everything should be in working order there.


README.md
```diff
@@ -118,6 +118,16 @@ If you rather have a question, please ask it on [our Discord server](https://dis
 
 We will always welcome pull requests on any of the [frctl organisation](https://github.com/frctl) repositories. Please submit PRs against `master` branch with an explanation of your intention.
 
+We use conventional commits, which means that every pull request title should conform to the standard.
```
LeBenLeBen, line 121:
```suggestion
We use [conventional commits](https://www.conventionalcommits.org/), which means that every pull request title should conform to the standard.
```

package.json
```diff
@@ -53,32 +41,26 @@
...
     "pre-commit": "^1.1.3",
     "sinon": "^9.0.1",
     "symlink-dir": "^4.0.1"
   },
-  "publishConfig": {
-    "access": "public"
+  "scripts": {
+    "clean": "lerna clean",
+    "setup": "lerna bootstrap --hoist --strict --ci --force-local",
+    "pretest": "symlink-dir ./packages/fractal node_modules/@frctl/fractal",
```
LeBenLeBen, line 60: I'm wondering if this should be changed to `npm run setup`. That’s basically what the bootstrap command does in addition with proper dependencies installation, no? It would also work with all the other packages we're going to add.

package.json
```diff
@@ -53,32 +41,26 @@
...
     "nyc": "^15.0.1",
     "pre-commit": "^1.1.3",
     "sinon": "^9.0.1",
     "symlink-dir": "^4.0.1"
   },
-  "publishConfig": {
-    "access": "public"
+  "scripts": {
+    "clean": "lerna clean",
+    "setup": "lerna bootstrap --hoist --strict --ci --force-local",
```
LeBenLeBen, line 59:
```suggestion
    "bootstrap": "lerna bootstrap --hoist --strict --ci --force-local",
```

How about naming this script `bootstrap` to be aligned with lerna’s?

README.md
```diff
@@ -118,6 +118,16 @@ If you rather have a question, please ask it on [our Discord server](https://dis
...
+
+### Development
+
+This repository is a monorepo managed by Lerna. There is only one lockfile in root. This means that all packages must be installed in root, manually added to the packages' package.json files and then bootstrapped with lerna.
+
+To do some work, run the following commands in root:
+1. `npm ci`
+2. `npm run setup`
```
LeBenLeBen, line 129:
```suggestion
2. `npm run bootstrap`
```

See below

LeBenLeBen:
@risker I think we should have a dedicated branch, either this one or another, to build up the whole monorepo, test everything together as we add package after package and only merge when everything is working as expected.

Then we can update the README of the imported repos to link here and archive them all at once.

package.json
risker: i was planning to refactor this out after importing mandelbrot, since in all seriousness @frctl/fractal should be re-introduced as mandelbrot devDependency since it very actually requires core.

risker:
Yeah, in that case I'd keep this one as the "temporary main monorepo branch" and import other packages to this one via PRs. In all cases, I can start working on importing other packages in separate branches, all that will change is the target branch :) I'll do my best to work on this during the weekend,

package.json
LeBenLeBen: 👍

risker:
I've added a todo list to the original post with a question regarding to what packages we want to import to the monorepo.

LeBenLeBen:
@risker I'd personally like to add `fractal-docs` within the monorepo as well. This will allow to change the docs within PRs directly and avoid having to create issues/PRs there when changes happens here.

If we don't want the docs to be updated before the release happen, we can use a different branch to deploy the site to Netlify. Even if I think we should rather be more specific in the docs and always specify the version in which something was introduce/changed/removed.

risker:
Agreed, although I think the docs repo needs some changes first - for example, it's the only package we have that uses yarn instead of npm lockfile. But I don't think that moving the docs repo is required in the MVP and can be imported later as well, as it does not affect other packages.

LeBenLeBen:
Ok sure, I created https://github.com/frctl/fractal-docs/issues/47 to tackle that. Please add a comment there if you other things that should be adapted comes to your mind.

risker:
Alright, all of the main repositories are now "import-ready". I would like to do the merging myself (as I've played the scenario through for the example branch) but I'd like someone to test it first and approve the PRs.

packages/fractal/package.json
```diff
@@ -0,0 +1,70 @@
...
+  "author": "Mark Perkins <mark@clearleft.com> (http://clearleft.com/)",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/frctl/fractal/issues"
+  },
+  "homepage": "https://github.com/frctl/fractal",
+  "dependencies": {
+    "@allmarkedup/fang": "^1.0.0",
+    "@frctl/handlebars": "^1.2.0",
+    "@frctl/mandelbrot": "^1.3.0",
```
LeBenLeBen, line 20: Should we update to 1.4.0 ?

risker: I updated the version in the mandelbrot import PR

PR #611, opened 2020-05-23
refactor: update fs-extra

### LINKED ISSUE #594: Update fs-extra dependency
This would upgrade a little bit of our legacy cruft that we need to eliminate:
- We are currently using fs-extra v0.30.0, which was released 4 years ago - in JavaScript years, that's like centuries ago.
- In v3.0.0, fs-extra added Promise support which we require - currently we "promisify" the old version through bluebird.

- Update fs-extra module
- Move to native promise-returning methods from fs-extra instead of "promisifying" sync methods

Modules that need refactoring:
- CLI new command: https://github.com/frctl/fractal/blob/master/src/cli/commands/new.js#L10
- Web Builder class: https://github.com/frctl/fractal/blob/master/src/web/builder.js#L8
- Web Theme class: https://github.com/frctl/fractal/blob/master/src/web/theme.js#L6 (I believe it's not used in that file, so we can just remove it)

mihkeleidast:
Resolves #594

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4267,6 +4267,11 @@
       "integrity": "sha1-x57Zf380y48robyXkLzDZkdLS3k=",
       "dev": true
     },
+    "at-least-node": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/at-least-node/-/at-least-node-1.0.0.tgz",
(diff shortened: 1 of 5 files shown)


LeBenLeBen:
Funny you opened that, I was about to open one to migrate from promisify(fs) to the new fs.promises (since Node 10). Maybe you want to add it to this PR? Otherwise I can do that separately.

mihkeleidast:
I'd rather keep this only for fs-extra and do the fs.promises changes in another PR :)

PR #612, opened 2020-05-23
chore: move EsLint/Prettier at root and fix all offenses

### LINKED ISSUE #590: Prettier config doesn’t match most of the repository existing style
There’s a `.prettierrc` file in the repository, which enables my editor auto-formatting on save. The problem is, the current style doesn't match Prettier rules; especially for quotes, Prettier wants doubles but the current code base use mostly singles.

This is an issue because when submitting PRs, a lot of changes happen that are unrelated to the fixes/feature proposed. I can live with "Save without formatting" for a while but that's neither convenient nor good for new contributors.

I think we should update the config to be as close as the existing code base style and then format all the files in a single commit. It should be done when there's almost no PRs open to avoid as many conflicts as possible.

LeBenLeBen:
Fixes #590


.eslintrc.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
+    parser: 'babel-eslint',
```
risker, line 2: not sure about this. the parser description says this:
> babel-eslint is a parser that allows ESLint to run on source code that is transformed by Babel.

but only the mandelbrot client-side theme code is transformed by Babel for us. i'm afraid if maybe this will allow new / experimental code that won't conform to our supported node versions.

.eslintrc.js
```diff
@@ -0,0 +1,21 @@
...
+    parserOptions: {
+        ecmaVersion: 6,
+    },
+    extends: ['eslint:recommended', 'plugin:prettier/recommended'],
+    env: {
+        browser: true,
+        node: true,
+        es6: true,
+    },
+    ignorePatterns: ['**/node_modules/**', '**/dist/**'],
```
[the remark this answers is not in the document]
LeBenLeBen: Done

.eslintrc.js
LeBenLeBen: Indeed, I moved this to Mandelbrot repository specifically. I changed the EcmaScript version to 2017 as it seems to be the lowest we can go to support all the current syntax and still be compatible with Node 10.

Issue #457, opened 2018-08-17
Add Integration tests for v1


dkhuntrods:
- [ ] Set up integration test framework (probably Cypress)
- [ ] Ensure Travis configured for:
  - [ ] OSX
  - [ ] Windows
  - [ ] Linux

- [ ] Add example projects:
  - [ ] (Default) handlebars
  - [ ] Nunjucks adapter
  - [ ] Twig adapter
  - ?

Chapabu:
I've had some thought around this that would be good to discuss. 

* It's a given that we're going to need a basic Fractal setup fixture with the tests, so that's fine. 
* OOTB, we're only going to be able to test Handlebars, as we'll want to test the other adapters on their own. 
* We also need to ensure we test things both when running as a server, and as a compiled instance (if possible). Running reliable tests with a built in server is not as straightforward as you'd want it to be, si I think focusing on compiled tests is a better place to start. We can add server tests later.
* As a next step, it'd be good to perhaps provide some kind of _simple_ abstraction so we can make writing tests for adapters a really simple thing

Chapabu:
Has any movement happened on this?

mikkelrom:
Not from my side at least, been two busy :(
Would it make any sense to add UI tests and/or end-to-end tests? These are the tests I've done the most :)

Chapabu:
The issue I see with UI/end-to-end tests, is that they'll _really_ be testing Mandlebrot for the most part. I mean...this is _useful_, but due to the nature of how Fractal is put together, I don't know:

1. Where these tests should go
2. Which repo these tests should be run against

Also, some other questions:

1. Do we expose some kind of helper system for testing other adapters (i.e. Nunjucks or Twig), or do we just test Handlebars as it's what we "ship" with.
2. Should we test the CLI processes? Like the build process?

mihkeleidast:
I want to close this since I don't see this as being directly actionable anymore. #615 added the framework for integration testing (for the default Handlebars adapter).

If we want to add integration tests the other adapters, those could (& IMO should) be filed as separate actionable issues and resolved as such. If so, they should probably primarily focus on the specialties of the adapter (filters, helpers, tags, etc), not the core parsing of the file system, which is covered by the work in #615. 

If anyone thinks there are actionable items left in this issue, please list them out so we could all understand what's left to be done.

PR #615, opened 2020-05-23
test: add Handlebars basic example

LeBenLeBen:
This is almost the default you get when running `fractal new`, I just added npm scripts. I plan on adding more components and pages to test more functionalities later and then duplicate it for Nunjucks and Twig adapter.

Relates to #457


packages/handlebars/package.json
```diff
@@ -21,6 +21,9 @@
     "promised-handlebars": "^2.0.1"
   },
   "peerDependencies": {
-    "@frctl/fractal": "^1.1.0"
+    "@frctl/fractal": "^1.3.0"
```
mihkeleidast, line 24: should not update the peerDependency version since we don't change any functionality

mihkeleidast:
And as discussed in Discord, let's add some actual tests with the example.

packages/handlebars/package.json
LeBenLeBen: Reverted

examples/handlebars/docs/01-index.md
```diff
@@ -0,0 +1,5 @@
+---
+title: Fractal default example
+---
+
+This is your index page. You can edit its contents at `docs/01-index.hbs`
```
mihkeleidast, line 5: the extension should be md here i think? is this a bug in our `new` command?

packages/handlebars/package.json
```diff
@@ -22,5 +22,8 @@
   },
   "peerDependencies": {
     "@frctl/fractal": "^1.1.0"
+  },
+  "devDependencies": {
```
mihkeleidast, line 26: can you rebase this on master? i added the devDeps to this file in #616, but in a slightly different location - just so this won't start conflicting

LeBenLeBen: Done

examples/handlebars/docs/01-index.md
LeBenLeBen: Yes, was a bug, fixed it.

examples/handlebars/components/notes/notes.spec.js
```diff
@@ -0,0 +1,23 @@
+const fractal = require('../../fractal.config.js');
+
+describe('tree', () => {
+    beforeEach(async () => {
+        await fractal.load();
+    });
+
+    it('loads notes from config', async () => {
```
mihkeleidast, line 8: we should also look if a variant with notes that's defined in the config gets the proper notes.

packages/fractal/views/cli/new/docs/index.hbs
```diff
@@ -2,4 +2,4 @@
 title: {{ projectTitle }}
```
mihkeleidast, line 2: this file itself should also be renamed to md extension

packages/handlebars/package.json
mihkeleidast: looks like this is still doubled

examples/handlebars/package.json
```diff
@@ -0,0 +1,20 @@
...
+    "test": "jest",
+    "test:dev": "npm run test -- --watch"
+  },
+  "dependencies": {
+    "@frctl/fractal": "^1.3.0",
+    "@frctl/handlebars": "^1.2.0",
+    "@frctl/mandelbrot": "^1.4.0"
+  },
+  "devDependencies": {
+    "jest": "^26.0.1",
```
Chapabu, line 17: I have a bit of an issue with this, but it's one that we may need to chat about in Slack.

We don't have many tests, but where we do - they're using Mocha/Chai/Sinon. The reason I'm not saying "Nope, we should change this" and "We should chat" instead, is that I'd REALLY like to migrate all our tests to Jest instead to reduce the amount of dependencies we need (as Jest literally covers everything we need).

mihkeleidast: I'm not against switching to Jest completely, but I'd wager that could be another issue/pr for another day. We don't exactly lose anything by having multiple test runners for a while. So if we could write this (migrating away from mocha/chai/sinon) up as an issue, I wouldn't block this PR.

Chapabu:
Got conflicts in the lockfile. Do we need to even commit these?

mihkeleidast:
> Got conflicts in the lockfile. Do we need to even commit these? They don't get published to NPM anyway, so they're only useful for contributors.

Yup, useful for contributors + for running tests predictably. No lockfile means that if a dependency breaks anything in a minor or patch, our tests will start failing. Of course, users will still get the breaking dependency versions, but the version diff will then be a good place to start looking for what went wrong.

examples/handlebars/package.json
LeBenLeBen: Yeah I'm also in favor of going full Jest, but I'd like to go step by step on this one as Mihkel said.

LeBenLeBen:
I should have time to tackle the remaining comments on Monday.

packages/handlebars/package.json
LeBenLeBen: Should be ok now

examples/handlebars/components/notes/notes.spec.js
LeBenLeBen: Done

packages/fractal/views/cli/new/docs/index.hbs
LeBenLeBen: Done

examples/README.md
```diff
@@ -0,0 +1,5 @@
+# Fractal examples
+
+This folder contains examples of Fractal instances using various adapters & settings.
+
+Although we use them for developing and testing Fractal, they are probably a great ressource for users as well.
```
mihkeleidast, line 5:
```suggestion
Although we use them for developing and testing Fractal, they are probably a great resource for users as well.
```

examples/handlebars/README.md
```diff
@@ -0,0 +1,33 @@
+# Fractal Handlebars example
+
+This is an example instance of Fractal using the default Handlebars adapter.
+
+## Setup
+
+[Follow these instructions](https://github.com/frctl/fractal#development) to setup the development environement, then run:
+
+```bash
+npm start
```
mihkeleidast, line 10: blabbed about this under the Mandelbrot PR, but we should somehow agree on a development process and figure out what the best way to get started is when getting the repo for the first time. Relevant point here is that before running the example instance, you need to build/start mandelbrot.

mihkeleidast: Cross-posting, since it's a bit relevant under both PRs:
> Found that running npx lerna run --scope=@frctl/handlebars-example --include-dependencies --parallel start for the example project will build mandelbrot (and any other dependencies that have a start script) too.

PR #616, opened 2020-05-23
style: add eslint-plugin-import

### LINKED ISSUE #613: Add eslint-plugin-import

In a Lerna monorepo with hoisted dependencies, forgetting to specify dependencies is pretty common. See https://github.com/lerna/lerna/blob/master/doc/hoist.md#forgetting-dependencies

Add [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import) with the recommended `no-extraneous-dependencies` rule turned on.

This will need a bunch of related changes as well, since we currently have a lot of package.json files that specify the main field. Since the rule looks at the closest package.json file, they get in the way. We can migrate away from those package.json files by instead adding index.js files that re-export the main code.

Note that this needs #612 to be merged first.

mihkeleidast:
Resolves #613


packages/handlebars/package.json
```diff
@@ -20,6 +20,9 @@
     "lodash": "^4.17.11",
     "promised-handlebars": "^2.0.1"
   },
+  "devDependencies": {
+    "@frctl/fractal": "^1.3.0"
+  },
   "peerDependencies": {
     "@frctl/fractal": "^1.1.0"
```
mikkelrom, line 27: This is probably not needed anymore then, right? 😊

packages/twig/package.json
```diff
@@ -18,6 +18,9 @@
     "lodash": "^4.17.11",
     "twig": "^1.14.0"
   },
+  "devDependencies": {
+    "@frctl/fractal": "^1.3.0"
+  },
   "peerDependencies": {
     "@frctl/fractal": "^1.1.0-beta.1"
```
mikkelrom, line 25: This is probably not needed anymore then, right? 😊

packages/handlebars/package.json
mihkeleidast: It is, since this tells the user that when installing `@frctl/handlebars`, core is also needed & the minimum required version is `1.1.0`.

Issue #614, opened 2020-05-23
Get rid of cycle warnings


mihkeleidast:
### What problem would this feature solve?

When developing this project and running `npm run bootstrap`, Lerna warns that we have dependency cycles. We should get rid of those cycles.

### What the feature should look like?

We have issues with the mandelbrot/fractal cycle, but we'll get more warnings when adapter dependencies are fixed after #613. To get rid of that, we need to create the following new packages:

- `@frctl/web` - this should include the whole [Web](https://github.com/frctl/fractal/tree/master/packages/fractal/src/web) structure.
- `@frctl/core` - this should include the whole [Core](https://github.com/frctl/fractal/tree/master/packages/fractal/src/core).

Once these parts are split from the main package, we should not have any remaining cycle warnings.

PR #621, opened 2020-05-25
refactor: split core into @frctl/core

mihkeleidast:
See #614 for background. I've tested this with #615 and it seemed to work okay.


packages/fractal/config.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
         version: null,
     },
     components: {
-        engine: '@frctl/handlebars',
+        engine: require('@frctl/handlebars'),
```
mihkeleidast, line 13: This is the only code change in this PR, it's because core tries to [require a string](https://github.com/frctl/fractal/blob/2341030b6c512e3483b07bdb259d9fdf76769594/packages/core/src/entities/source.js#L40) by itself and that fails since in the context of this repository `@frctl/handlebars` does not exist in the node_modules tree of `@frctl/core`. It should not matter for real uses, however, since in that case core and handlebars would always exist in the same node_modules directory.

LeBenLeBen:
@Chapabu I’m not really comfortable reviewing this, I guess you’d give better inputs than me.

PR #623, opened 2020-05-26
refactor: change bin file to js & lint/format

mihkeleidast:
Since the bin file did not have a js extension, it was not linted/formatted in #612. This also means that it currently bypasses all import lint rules that are rather important.

Looking at the JS ecosystem, it seems fairly common that the bin file is a JS file, so I'm fairly comfortable changing this.

### DIFF
--- a/packages/fractal/bin/fractal.js
+++ b/packages/fractal/bin/fractal.js
@@ -2,17 +2,15 @@
 
 'use strict';
 
-const Path           = require('path');
-const semver         = require('semver');
-const Liftoff        = require('liftoff');
-const chalk          = require('chalk');
+const Path = require('path');
+const semver = require('semver');
+const Liftoff = require('liftoff');
+const chalk = require('chalk');
 const updateNofifier = require('update-notifier');
-const cliPackage     = require('../package.json');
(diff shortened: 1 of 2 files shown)


packages/fractal/bin/fractal.js
```diff
@@ -40,10 +38,11 @@ try {
     if (projectPackage.fractal && projectPackage.fractal.main) {
         config.configPath = Path.join(process.cwd(), projectPackage.fractal.main);
     }
-} catch(e){}
-
-FractalCli.launch(config, function(env){
+} catch (e) {
+    console.log(e);
```
Chapabu, line 42: I'd make this `console.error(e)` as it's literally logging the error.

packages/fractal/bin/fractal.js
```diff
@@ -53,7 +52,7 @@ FractalCli.launch(config, function(env){
         try {
             app = require(configPath);
             scope = 'project';
-        } catch(e) {
+        } catch (e) {
             console.log(e.stack);
```
Chapabu, line 56: Same here re: `console.error`

PR #627, opened 2020-05-29
feat: allow customizing skin with hex codes

### LINKED ISSUE #601: Customizing theme using hex-codes
I've read the documentation on customizing the theme but I didn't see a way of customizing the theme's color using hex-codes, rather than using one of the built in color schemes.

mihkeleidast:
Resolves #601.

Following this change, users will be able to customize Mandelbrot's sking with hex codes.

Previously we allowed specifying only the skin name:
```js
const mandelbrot = require('@frctl/mandelbrot')({
    skin: 'fuchsia',
});
```

Now we will allow specifying an object with skin name and hex codes:
```js
const mandelbrot = require('@frctl/mandelbrot')({
    skin: {
        name: 'default',
        accent: '#f2d',
        complement: '#402',
        links: '#a06',
    },
});
```
We'll allow three colors - following the same naming convention in our skins definition. If the user's browser does not support CSS variables, it'll use the fallback color scheme (which defaults to default).

This will need documenting, once merged.

### Technical gibberish
- I'm converting the user-specified hex codes into RGB for the CSS variables. This is so we could use rgba in our styles, where needed. [See more](https://stackoverflow.com/questions/29591465/use-css-variables-with-rgba-for-gradient-transparency)
- All of the CSS variables are defined as rgb/rgba in a local variables. This is because Sass otherwise interprets the rgb/rgba as its own functions and it won't work. [See more](https://github.com/sass/node-sass/issues/2251)

### DIFF
--- a/packages/mandelbrot/assets/scss/core/_foundation.scss
+++ b/packages/mandelbrot/assets/scss/core/_foundation.scss
@@ -5,7 +5,10 @@
 }
 
 ::selection {
+    --background-color: rgba(var(--skin-links), 0.2);
+
     background-color: rgba($color-link, 0.2);
+    background-color: var(--background-color, rgba($color-link, 0.2));
 }
 
 html {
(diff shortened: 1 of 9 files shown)


packages/mandelbrot/assets/scss/core/_foundation.scss
```diff
@@ -70,15 +73,22 @@ ul {
...
+    color: var(--color, rgba($color-link, 0.9));
     transition: all 0.15s ease-out;
 
     &:active {
         color: $color-link;
+        color: var(--color, rgba($color-link, 0.9));
     }
 
     &:hover {
-        color: darken($color-link, 15%);
```
mihkeleidast, line 81: i changed the logic here because if we want to use darken with css variables, i'd need to use hsl everywhere and that would be even more complicated code-wise. using rgba everywhere is just simpler.

PR #628, opened 2020-05-30
feat: allow overriding/extending highlighter

### LINKED ISSUE #582: Allow Highlight.js language configuration within fractal config file
It would be ideal to be able to install `highlight.js` (via `npm i highlight.js`), define custom language and configuration and then pass that configuration to fractal. At the moment, it doesn't seem to be possible to overwrite fractal highlight configuration and add custom language.

Something like the following would be nice :
```
const hljs = require('highlight.js');
hljs.registerLanguage('vue', () => ({ ... }));
... 
const fractal = module.exports = require('@frctl/fractal').create();
fractal.components.highlight(hljs);
```

mihkeleidast:
Resolves #582.

For example, adding svelte language support:
```js
const hljs = require('highlight.js');
const hljs_svelte = require('highlightjs-svelte');
const _ = require('lodash');

hljs_svelte(hljs);

fractal.web.set('highlighter', function (content, lang) {
    content = _.toString(content || '');
    lang = lang ? lang.toLowerCase() : lang;
    try {
        return lang ? hljs.highlight(lang, content).value : hljs.highlightAuto(content).value;
    } catch (e) {
        return hljs.highlightAuto(content).value;
    }
});
```
I thought about just allowing passing in an extended HighlightJS, but this would be more dynamic - it would allow to use some other highlighting lib if necessary.

Needs documenting after merge.

### DIFF
--- a/packages/fractal/config.js
+++ b/packages/fractal/config.js
@@ -128,5 +128,6 @@ module.exports = {
         assets: {
             mount: 'assets',
         },
+        highlighter: require('@frctl/core').highlighter,
     },
 };
--- a/packages/web/src/engine/filters/highlight.js
+++ b/packages/web/src/engine/filters/highlight.js
@@ -1,10 +1,8 @@
 'use strict';
 
-const highlight = require('@frctl/core').highlighter;
-
-module.exports = function () {
+module.exports = function (app) {
     return {
         name: 'highlight',
-        filter: (str, lang) => highlight(str, lang),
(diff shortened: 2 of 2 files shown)


mihkeleidast:
There was an idea during the monorepo move that maybe we should move the docs into the monorepo as well - then we could add docs with the same PR. But that needs a bit more thinking, since we need to have a process to mark features with versions they were released with etc.

But that action would be sweet as well and it shouldn't be too complicated really!

PR #631, opened 2020-06-01
docs: document how to work on Mandelbrot

### LINKED ISSUE #609: Need to document how to clone and build the theme

LeBenLeBen:
fixes #609

Should be merged after #615 as it references it.

### DIFF
--- a/packages/mandelbrot/README.md
+++ b/packages/mandelbrot/README.md
@@ -2,22 +2,33 @@
 
 The default web UI theme for Fractal.
 
-See the Fractal [documentation](http://fractal.build/guide) for details on configuration and usage.
+[![NPM Version](https://img.shields.io/npm/v/@frctl/mandelbrot.svg?style=flat-square)](https://www.npmjs.com/package/@frctl/mandelbrot)
+
+See the Fractal [documentation](https://fractal.build/guide) for details on [configuration and usage](https://fractal.build/guide/web/default-theme.html).
 
 In order to get a locale aware date for the 'last updated' text, install the
 [full-icu](https://github.com/unicode-org/full-icu-npm) module.
 
-## Customize theme labels
+## Development
 
-Some theme-specific labels can be overridden via config options:
+[Follow these instructions](https://github.com/frctl/fractal#development) to setup the development environement, then go to the Mandelbrot package directory:
 
 ```
-const theme = require('@frctl/mandelbrot')({
-    labels: {
-        info: 'Information',
-        builtOn: 'Built on',
-    },
-});
-
-fractal.web.theme(theme);
+cd packages/mandelbrot
 ```
+
+In there, run:
+
+```
+npm start
+```
+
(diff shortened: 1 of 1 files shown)


packages/mandelbrot/README.md
```diff
@@ -2,22 +2,33 @@
...
 In order to get a locale aware date for the 'last updated' text, install the
 [full-icu](https://github.com/unicode-org/full-icu-npm) module.
 
-## Customize theme labels
+## Development
+
+[Follow these instructions](https://github.com/frctl/fractal#development) to setup the development environement, then go to the Mandelbrot package directory:
+
+```
+cd packages/mandelbrot
```
mihkeleidast, line 17: this certainly works, but since we're in a lerna monorepo, all commands can be run from root as well, e.g. `npx lerna exec --scope=@frctl/mandelbrot -- npm run start`. I'm not exactly sure what the "official" approach is when developing in a monorepo, but I'm pretty used to running everything from root.

In general, I think that if going forward we're still going to have the single official theme, we could also add a shortcut script to root (for example `npm run mandelbrot -- npm start`) that would just alias the longer lerna command.

Since developing the theme requires to run both the mandelbrot and a test example (if they're not linking it to their own instance of fractal), we can do `npx lerna exec --scope=@frctl/mandelbrot --scope=@frctl/handlebars-example -- npm start` to start both dev servers.

Don't change anything based on (only) this reply though - just starting a conversation here, I haven't really figured out how the dev process should look like. For example, if you don't want to develop mandelbrot, but rather tinker around with the handlebars-example, you still need to build mandelbrot first. And what I've personally experienced a lot is that if you want to develop an adapter or the core module, it's a bunch of start/stop/start/stop for node to pick up changes to source files after every change to a console.log. Maybe I've just been doing it wrong? Or is there a more comfortable way to develop a Node app such as Fractal?

packages/mandelbrot/README.md
```diff
@@ -2,22 +2,33 @@
...
+```
 
-Some theme-specific labels can be overridden via config options:
+In there, run:
 
 ```
-const theme = require('@frctl/mandelbrot')({
-    labels: {
```
mihkeleidast, line 16: I think this is not documented on the docs site, maybe open an issue there if we're removing this info from here?

packages/mandelbrot/README.md
mihkeleidast: Found that running `npx lerna run --scope=@frctl/handlebars-example --include-dependencies --parallel start` for the example project will build mandelbrot (and any other dependencies that have a start script) too.

LeBenLeBen: Opened https://github.com/frctl/fractal/issues/632 to tackle this separately

packages/mandelbrot/README.md
LeBenLeBen: https://github.com/frctl/fractal-docs/issues/57

PR #642, opened 2020-06-20
feat: allow pulling collections to root

### LINKED ISSUE #103: Feature request: Allow for multiple root level ‘component’ folders
Within my components folder, I have a set of components which are top-level (page) assemblies. _Conceptually_, these are still components, but in terms of my deliverable, they are markedly different. Currently I have made the distinction in the navigation by using italics:

screen shot 2016-07-28 at 10 51 56

Ideally, the Web UI would allow me to place them under a top level heading instead, i.e.:

screen shot 2016-07-28 at 10 51 22

This might be a Mandelbrot issue (and one that could be solved by using a custom theme), but it’s worth pointing out that allowing only a single component folder is a notable opinion within Fractal.

mihkeleidast:
Resolves #103.

### DIFF
--- a/packages/mandelbrot/views/macros/navigation.nunj
+++ b/packages/mandelbrot/views/macros/navigation.nunj
@@ -19,8 +19,7 @@
     {% for item in items %}
         {% if item.isCollection or (item.isComponent and not item.isCollated and item.variants().filter('isHidden', false).size > 1) %}
             {# skip root collections because they are pulled out of the tree #}
-            {% if item.isCollection and item.isRoot %}
-            {% else %}
+            {% if not (item.isCollection and item.isRoot) %}
             <li class="Tree-item Tree-collection Tree-depth-{{ depth }}" data-behaviour="collection" id="tree-{{ root.name }}-collection-{{ item.handle }}">
                 <button class="Tree-collectionLabel" data-role="toggle" aria-expanded="true" aria-controls="tree-{{ root.name }}-collection-{{ item.handle }}-items">
                     {{ item.label }}
--- a/packages/mandelbrot/views/macros/navigation.nunj
+++ b/packages/mandelbrot/views/macros/navigation.nunj
@@ -18,19 +18,23 @@
 {% macro leaves(items, root, current, depth, request) %}
     {% for item in items %}
         {% if item.isCollection or (item.isComponent and not item.isCollated and item.variants().filter('isHidden', false).size > 1) %}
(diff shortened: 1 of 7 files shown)


packages/mandelbrot/views/macros/navigation.nunj
```diff
@@ -18,19 +18,23 @@
...
         {% if item.isCollection or (item.isComponent and not item.isCollated and item.variants().filter('isHidden', false).size > 1) %}
-        <li class="Tree-item Tree-collection Tree-depth-{{ depth }}" data-behaviour="collection" id="tree-{{ root.name }}-collection-{{ item.handle }}">
-            <button class="Tree-collectionLabel" data-role="toggle" aria-expanded="true" aria-controls="tree-{{ root.name }}-collection-{{ item.handle }}-items">
-                {{ item.label }}
-            </button>
-            <ul class="Tree-collectionItems" data-role="items" id="tree-{{ root.name }}-collection-{{ item.handle }}-items">
-            {% if item.isComponent and not item.isCollated %}
-            {% set items = item.variants().filter('isHidden', false).items() %}
+            {# skip root collections because they are pulled out of the tree #}
+            {% if item.isCollection and item.isRoot %}
```
LeBenLeBen, line 22: I think we can avoid the empty block with this condition

```suggestion
            {% if not (item.isCollection and item.isRoot) %}
```

packages/mandelbrot/views/partials/navigation/components.nunj
```diff
@@ -1,5 +1,13 @@
 {% if frctl.components.size %}
-<div class="Navigation-group">
-{{ nav.tree(frctl.components, entity, request) }}
-</div>
+{% set rootCollections = frctl.components.rootCollections().filter('isHidden', false) %}
+{% if rootCollections.size > 0 %}
```
LeBenLeBen, line 3: This isn’t really useful as we’ll simply loop on nothing on the next line if there’s no collection

packages/mandelbrot/views/partials/navigation/components.nunj
```diff
@@ -1,5 +1,13 @@
 {% if frctl.components.size %}
```
LeBenLeBen, line 1: This condition is not really necessary anymore as we are checking that a collection includes entities before displaying it

LeBenLeBen (approved):

packages/mandelbrot/views/macros/navigation.nunj
mihkeleidast: not kidding when i say i spent a lot of time trying to figure out an optimal way to negate this condition - the answer just did not come to me :)

PR #646, opened 2020-08-11
fix: resolve the context for fully rendered subcomponents

justafish:
Follow up from #573 that resolves the context correctly

Steps to reproduce:
- Create a chain of components that use other pre-rendered components e.g.

``` link.config.yml
label: Links
context:
  text: "Link"
  href: "#"
```
``` link.twig
<a href="{{ href }}">{{text}}</a>
```

``` menu.config.yml
label: Menu
context:
  links:
    - "@@link"
    - "@@link"
    - "@@link"
    - "@@link"
```
``` menu.twig
<ul>
  {% for link in link %}
    <li>{{ link }}</li>
  {% endfor %}
</ul>
```

``` navigation.config.tml
label: Navigation
context:
  menu: "@@menu"
```
``` navigation.twig
<nav>
{{ menu }}
</nav>
```

- Open the Navigation component and see that instead of the links being rendered, you see `@@link @@link @@link @@link`
- Apply this PR and see the links are now rendered


mihkeleidast:
This is nice! Would you be up for adding a few integration tests for this functionality under the handlebars example package?

justafish:
@mihkeleidast I've added a test but it doesn't pass for handlebars as I've only fixed this for Twig. I'm not at all familiar with the handlebars or nunjucks adapters, so I'm not sure what the equivalent fixes would be. Can you give any guidance?

examples/handlebars/components/context/context.spec.js
```diff
@@ -34,4 +34,9 @@ describe('tree', () => {
         const ctx = await fractal.components.find('@reference-key').getResolvedContext();
         expect(ctx.text).toBe('JS Context');
     });
+
+    it('loads fully rendered component from another component context key', async () => {
+        const ctx = await fractal.components.find('@reference-full').getResolvedContext();
+        expect(ctx.text).toBe('JS Context');
```
mihkeleidast, line 40: shouldn't this assert `ctx.reference`?

mihkeleidast (commented):
I don't actually get why this needs a fix in adapters, since everything else to support this feature is in the resolver. What's the context for that fix?

examples/handlebars/components/context/context.spec.js
justafish: :woman_facepalming: 
thank you!

justafish:
Thanks @mihkeleidast the tests are passing now :tada:

packages/twig/src/adapter.js
```diff
@@ -22,6 +22,10 @@ class TwigAdapter extends Fractal.Adapter {
              */
 
             Twig.Templates.registerLoader('fractal', function (location, params) {
+                if (self.views.length === 0) {
```
mihkeleidast, line 25: what does this change do? it feels wrong here, e.g. it does not seem like a specific adapter issue. maybe this should be fixed upstream in the Adapter class or somewhere else entirely? I can't suggest something if I don't understand what behavior this changes though :)

justafish: @mihkeleidast When fractal is started it'll sometimes throw errors for referenced components because it's trying to resolve paths for views it hasn't yet loaded e.g.
```
[0] ✘ Error: Error: Template @component/widget/icon-link/icon-link.twig not found
```

As the tests are passing now I think this is an issue specific to the twig adapter

mihkeleidast: How can this be reproduced? How is it related to the `@@component` resolving syntax? This feels like a separate issue / fix and out of place here.

mihkeleidast: possible alternative solution (might not work as I don't understand the issue completely) that would go in the resolver: on line 58 (e.g. for the regular context resolver) there are if-clauses that check if components exist etc. shouldn't there be some if clauses around the context renderer on line 43?

justafish: Here's a minimal example demonstrating the issue: https://github.com/justafish/test-fractal
```
yarn install
yarn start
test-fractal on  main [!] is 📦 v1.0.0 via ⬢ v14.11.0 took 22s 
❯ yarn start
yarn run v1.21.1
$ fractal start --sync
✘ Error: Error: Template @component/widget/icon-link/icon-link.twig not found
┌──────────────────────────────────────────┐
│ Fractal web UI server is running!        │
│──────────────────────────────────────────│

```

re alternative solution - I did look into that but tbh I can't recall why it wasn't working out :smile: but I do think it's related to how the twig adapter is loading templates vs handlebars (and hence the tests for that passing fine)

mihkeleidast: Thanks, I'll try to look at it tomorrow.

mihkeleidast: Debugged this a bit, it seems like a deeper issue somewhere in Fractal core.

Not sure how or why, but in your fractal.config.js if you set the engines like this:
```
fractal.docs.set('engine', twigAdapter);
fractal.components.set('engine', twigAdapter);
```
everything works correctly as far as I can understand.

As far as I can understand, setting the engine by method [forces an early load here](https://github.com/frctl/fractal/blob/main/packages/core/src/entities/source.js#L55), and Fractal has not actually parsed the file system at that stage yet (verified this by adding a log [here](https://github.com/frctl/fractal/blob/main/packages/fractal/src/fractal.js#L104)). If you set the config property, Fractal will load the engine a bit later and the views are loaded by then.

I have no idea why it works this way as it's been documented forever to register the engine via the method. I'll discuss this with my peers and maybe we can figure out what is going on here.

justafish: oh wow, yep that worked :exploding_head: 

I've removed the twig change from this PR in the meantime so this can move forward at least

mihkeleidast (approved):

PR #652, opened 2020-08-25
fix(fractal): fix resolving camelCased variant templates

### LINKED ISSUE #651: In configuration the filename of view ignores letter case

1. Create a component with a variant.
1. Create a template file for the variant using camel case (e.g. `myComponent--templateForVariant.html`).
1. In the configuration of the variant specify the view template created (e.g. `view: 'myComponent--templateForVariant.html'`).

**Example for configuration:**
```
variants:
  -
    name: variant
    label: My Variant
    view: 'myComponent--templateForVariant.html'
````

### Result

Instead of the template of the variant the template of the component itself will be used for rendering.

### Expected behaviour

The template with the given filename should be used.

When using only lowercase in the configuration by leave the filename in camel case all works fine. By the way this is not the only issue when using file or component names with camel case.

### Context

-   Fractal version: 1.1.7
-   Node version: 10.16.3

### PR DESCRIPTION
**Pull request by mihkeleidast**
Fixes #651.

### DIFF
--- a/examples/handlebars/components/render/render.spec.js
+++ b/examples/handlebars/components/render/render.spec.js
@@ -1,0 +1,32 @@
+const fractal = require('../../fractal.config.js');
+
+describe('render', () => {
+    beforeEach(async () => {
+        await fractal.load();
+    });
+
+    it('renders default template for default variant', async () => {
+        const render = await fractal.components.find('@render').render();
+        expect(render).toBe('Render something\n');
+    });
+
+    it('renders default template for regular variant', async () => {
+        const render = await fractal.components.find('@render--variant-1').render();
+        expect(render).toBe('Render something\n');
+    });
+
+    it('renders variant template for variant with the same name', async () => {
+        const render = await fractal.components.find('@render--variant-2').render();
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: examples/handlebars/components/render/render.spec.js [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+    });
+
+    it('renders default template for default variant', async () => {
+        const render = await fractal.components.find('@render').render();
+        expect(render).toBe('Render something\n');
+    });
+
+    it('renders default template for regular variant', async () => {
+        const render = await fractal.components.find('@render--variant-1').render();
+        expect(render).toBe('Render something\n');
```
LeBenLeBen, line 15: I think we should change the content of this variant to make it unique and thus prevent any future bug where the template used would be wrong but the test would still pass.

LeBenLeBen (changes requested):

examples/handlebars/components/render/render.spec.js
mihkeleidast: Agreed!

LeBenLeBen (approved):

PR #654, opened 2020-08-26
ci: add node 14 to test matrix

mihkeleidast:
I wouldn't merge this as of yet, I only added this to see if our current tests pass or what issues may arise. I'd add it officially when 14 will be set to LTS status.

Edit: I totally expected this to fail at some things in tests, but apparently not?


mihkeleidast:
Closing this in favor of #662.

PR #659, opened 2020-09-08
fix(mandelbrot): make relative path links from notes accessible both from local server as deployed server

### LINKED ISSUE #658: Relative path links from notes accessible both from local server as deployed server.

Possibility to access components via the same relative path (with extension) between the local development server and the build variant being hosted elsewhere.

Right now, if you put a relative URL in the notes pointing to another component, for instance: `./component-name`, it will work fine on the local development server. The default settings on the builder make it work so it contains the extension (.html). Therefore when you host the build variant elsewhere, your notes with the `./component-name` the link will result in a 404.

Vice versa this has the same effect, so if we would construct the link to contain the extension, as such: `./component-name.html`, this will work fine in the build variant. But this won't work in the local development server as it will try and match the handle including the extension (stating: Not found - Component 'component-name.html') not found.

It would be a bit more decent, functionality-wise, if these inner links don't break depending on the environment its on.

A possible solution is to overwrite the default extension the builder script receives from the config, and change the htaccess file to point towards the extension-less build files, this is already implemented. But, as is the case in my current environment, I don't have the opportunity to go and modify that. Therefore, why not have the possibility to inverse this mechanism from the local environment as well?

From the config, it could supply an extension (by default '.html') which then gets retrieved in the local server and somehow knows how to match the appropriate component. That way he will strip off the extension and can match the handle accordingly.

**Proposed solution (direction)**: #659 - see discussion below on cleanliness

ccGabbles:
Fixes #658 

Path Helper conditional extended as readme.md files don't send the environment on the same location in the root object.

### DIFF
--- a/packages/fractal/config.js
+++ b/packages/fractal/config.js
@@ -108,6 +108,7 @@ module.exports = {
             sync: false,
             watch: false,
             port: null,
+            ext: '.html',
             syncOptions: {},
         },
         builder: {
--- a/packages/fractal/config.js
+++ b/packages/fractal/config.js
@@ -108,7 +108,6 @@ module.exports = {
             sync: false,
             watch: false,
             port: null,
-            ext: '.html',
             syncOptions: {},
         },
         builder: {
--- a/packages/handlebars/src/helpers/path.js
+++ b/packages/handlebars/src/helpers/path.js
@@ -21,7 +21,7 @@ module.exports = function (fractal) {
             path = tpl(context);
         }
 
-        if (!root || !root._env || root._env.server) {
+        if (!root || !root._env || (root._env.server || root.server)) {
             return path;
         }
 
--- a/packages/handlebars/src/helpers/path.js
+++ b/packages/handlebars/src/helpers/path.js
@@ -21,7 +21,7 @@ module.exports = function (fractal) {
             path = tpl(context);
         }
 
-        if (!root || !root._env || (root._env.server || root.server)) {
+        if (!root || !root._env || root._env.server) {
             return path;
         }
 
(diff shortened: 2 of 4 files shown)


mihkeleidast:
I think the actual bug is in mandelbrot [here](https://github.com/frctl/fractal/blob/main/packages/mandelbrot/views/partials/browser/panel-notes.nunj#L4) - the docs.renderString method should accept env as it's third parameter, not second, as seen [here](https://github.com/frctl/fractal/blob/main/packages/fractal/src/api/docs/source.js#L83). Pretty sure if that renderString call is changed to `frctl.docs.renderString(entity.notes, null, renderEnv)`, this bug will be squashed and nothing else will break. In other places where docs are rendered in mandelbrot, context is passed as `null` apparently - see [here](https://github.com/frctl/fractal/blob/main/packages/mandelbrot/views/pages/doc.nunj#L21) for example.

mihkeleidast (changes requested):
Let's fix this from mandelbrot

ccGabbles:
Changed the code accordingly, indeed makes more sense to apply the change in the proposed spot. Verified the change in de repro project and works there, no regression issues found either. Thank you for thinking with me here! :)

ccGabbles:
@mihkeleidast, fixed as proposed, do you have time to take a look at it again? :)

mihkeleidast:
I'll try to take a look sometime this week, been busy with moving & that cuts into my time usually allocated for Fractal.

mihkeleidast (approved):

mihkeleidast:
Thanks! Sorry it took a bit long to review.

PR #661, opened 2020-10-07
feat: filter excluded files/directories in the file system parsing stage

### LINKED ISSUE #660: Exclude option is not taken into account when parsing the file tree.

The exclude option was added in #572. It successfully discards excluded paths from the component tree. However, it's currently implemented in a way that the exclusion is done when fractal is creating the component tree from the file system. This means that Fractal still parses all of those excluded paths into it's representation of directory structure, which makes the whole process quite slow when there are a lot of excluded directories. It's also important to point out that Fractal currently re-parses the entire file tree when any of the files have changed, so this makes reloads very slow when trying to work inside Fractal.

I POC-ed a solution and measured the impacts a bit. Currently, the parsing of the file tree, parsing of the components and resolving the context takes about 38 seconds in my (huuuge) project:
image

With a POC fix applied, the process takes about 6 seconds. That's a 6x improvement.
image

The files / folders should be excluded in the file system parsing stage.

I'll attach my POC solution here since I tested this out by debugging in my project's node_modules :) It needs to be fully reworked since the Fractal codebase has changed a bit since last release.

fractal/src/core/mixins/source.js
[poc.zip](https://github.com/frctl/fractal/files/5314211/poc.zip)

mihkeleidast:
Fixes #660. 

This will increase file system parsing performance when there are huge amounts of excluded files and/or directories. For example. in my project (which is a monorepo where there are a lot of excluded node_modules dirs), I counted how many times the [build function](https://github.com/frctl/fractal/blob/main/packages/core/src/fs.js#L52) was called. Before this change, it was called 30412 times. After this change, only 2269 times. Time-wise it's something like 36s down to 6s. So it's pretty obvious how much perf this gives us, since Fractal parses the entire tree on every file change.

### DIFF
--- a/packages/core/src/mixins/source.js
+++ b/packages/core/src/mixins/source.js
@@ -202,24 +202,18 @@ module.exports = mixin(
                 const exclude = this.get('exclude');
                 const matchers = [];
 
-                function addMatcher(matcher) {
-                    if (_.isString(matcher)) {
-                        matchers.push(`${matcher}`);
-                    }
-                }
-
                 if (_.isArray(exclude)) {
                     for (const excludeItem of exclude) {
-                        addMatcher(`${excludeItem}`);
+                        matchers.push(`${excludeItem}`);
                     }
                 } else {
-                    addMatcher(`${exclude}`);
+                    matchers.push(`${exclude}`);
                 }
 
                 return fs.describe(
                     this.fullPath,
                     this.relPath,
-                    exclude ? (filePath) => !anymatch(matchers, filePath) : undefined,
+                    matchers.length ? (filePath) => !anymatch(matchers, filePath) : undefined,
                     this.get('ext')
                 );
             }
--- a/packages/core/src/mixins/source.js
+++ b/packages/core/src/mixins/source.js
@@ -199,7 +199,29 @@ module.exports = mixin(
             }
 
             _getTree() {
-                return fs.describe(this.fullPath, this.relPath, undefined, this.get('ext'));
+                const exclude = this.get('exclude');
+                const matchers = [];
+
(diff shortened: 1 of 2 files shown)


packages/core/src/mixins/source.js
```diff
@@ -199,7 +199,29 @@ module.exports = mixin(
...
 
             _getTree() {
-                return fs.describe(this.fullPath, this.relPath, undefined, this.get('ext'));
+                const exclude = this.get('exclude');
+                const matchers = [];
+
+                function addMatcher(matcher) {
+                    if (_.isString(matcher)) {
+                        matchers.push(`${matcher}`);
+                    }
```
LeBenLeBen, line 208: This is weird, and I know it was like that before, but still 😅 We check if it’s a string but we always give it a string anyway, no? I guess we could remove this whole function and directly push on line 213 and 216, wdyt?

mihkeleidast: hm, we could do that. i guess the check is here because the fractal config is not very type-safe, e.g. users can set almost any types of values for this option. but we don't really want to allow object etc to be added as matchers?

then again, anymatch allows functions, regex etc. so this check does not actually make any sense.

mihkeleidast: Oh, looked at the previous code and now i get it - previously we needed to negate the matchers by prepending `!` so it had to be a string. but now we don't need the negation anymore, so we can allow non-string matchers as well.

actually we don't even need the array/push anymore, we can send it directly to anymatch :)

LeBenLeBen: Nice!

LeBenLeBen (approved):

PR #666, opened 2020-10-15
docs: update, streamline and add missing readmes

mihkeleidast:


packages/core/README.md
```diff
@@ -0,0 +1,9 @@
+# @frctl/core
+
+The core module of [Fractal](http://github.com/frctl/fractal).
```
LeBenLeBen, line 3: It would be great to shortly describe what core and web modules are dedicated to, that would help some to understand the architecture better.

LeBenLeBen (changes requested):

packages/core/README.md
mihkeleidast: Agreed, but there's no clear separation / idea behind it right now except that they're modular enough so that lerna would not give us circular dependency warnings :) And honestly I could not describe the intent of these modules right now even if I wanted to - there needs to be a whole lot work done before I can get my head wrapped around how Fractal has been built and _why_ exactly this way.

README.md
```diff
@@ -11,26 +11,17 @@
 
 <br />
 <div align="center">
-  <!-- Travis -->
-  <a href="https://travis-ci.org/frctl/fractal">
-    <img src="https://img.shields.io/travis/frctl/fractal?label=travis%20ci" alt="">
-  </a>
+
+  <!-- Github Actions -->
```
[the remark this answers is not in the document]
mihkeleidast: Ah that's why the badges were not using the markdown syntax & instead used inline html... I'll revert that part.

LeBenLeBen (approved):

PR #669, opened 2020-10-17
fix(Mandelbrot): Properly initialize assets tab file select

### LINKED ISSUE #668: Mandelbrot: assets tab select2 throw an error on initialization
```
﻿TypeError: $(...).select2 is not a function
```

LeBenLeBen:
Closes #668

Select2 changed the way it detects modules starting 4.0.4, see [this comments thread](https://github.com/select2/select2/commit/45a877345482956021161203ac789c25f40a7d5e). Apparently it’s still not fixed in the latest version 4.0.13 so let’s go for the workaround.

FTR the regression on our side was introduced since [this commit](https://github.com/frctl/fractal/commit/e2fe7699dc47b1177894fc85c57f2e6d16fdf71f#diff-b3c812e431a045641b9e7129190c5ff4b31bab06bda461695572ad1e079b3f6eL55).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4380,12 +4380,6 @@
         }
       }
     },
-    "almond": {
-      "version": "0.3.3",
-      "resolved": "https://registry.npmjs.org/almond/-/almond-0.3.3.tgz",
-      "integrity": "sha1-oOfJWsdiTWQXtElLHmi/9pMWiiA=",
-      "dev": true
-    },
     "amdefine": {
       "version": "1.0.1",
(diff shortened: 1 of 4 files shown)


mihkeleidast:
Shouldn't the lockfile in root update right now? Looks like it would still install select2@4.0.3 right now if using `npm ci`

LeBenLeBen:
@mihkeleidast Yeah my bad, I mixed things up before commit, should be good now.

mihkeleidast (approved):

Issue #647, opened 2020-08-13
Optionally show custom date format in Information "Built on"


nessthehero:
### What problem would this feature solve?

Would give someone reviewing the component library an idea of when it was exactly compiled, especially if an ongoing quality assurance test is in progress.

### What the feature should look like?

Either with a new property passed to the theme in the fractal config to enable the `H:m:s`, or the ability to pass a date format into the theme config to get a more precise display.

```
const myCustomisedTheme = mandelbrot({
	skin: "black",
        date_format: 'F j, Y, g:i a' // August 13, 2020 9:23 am
});
```

PR #678, opened 2020-10-25
feat: allow custom information properties

### LINKED ISSUE #624: Allow custom information properties

Showing the semantic version out of package.json beside of the information when it was last built.

Allow passing an object with freely defined key value pairs that will be shown together with `builtOn`. I would also think of replacing this with `builtOn` entirely (or only show it when the custom option is empty) since this would allow passing a custom `builtOn` with custom date formatter, e.g. "01.01.2020" instead of "01/01/2020".

Example:

```js
fractal.web.theme(mandelbrot({
  skin: 'navy',
  panels: ['html', 'view', 'context', 'resources', 'notes'],
  favicon: '/images/common/favicon.ico',
  information: {
    'Version': pkg.version
  }
}));
```

For now I'm going to show the version in the project title:

```js
fractal.set('project.title', `${pkg.description} <b><i style="color:#80a753">v${pkg.version}</i></b>`);
```

mihkeleidast:
Resolves #624 and #647.

This will allow users to customize the information properties like this:
```
const mandelbrot = require('@frctl/mandelbrot');

fractal.web.theme(
    mandelbrot({
        information: [
            {
                label: 'version',
                value: '1.2.3',
            },
        ],
    })
);
```
An information property's type can be set to `time` (defaults to `string`), in which case the property is marked up as `<time>` element.
Additionally, we provide an optional format prop if the base value needs to be formatted to a specific output - the usecase here is allowing to customize the date format, but I guess there could be other uses as well, if we should add other property types in the future.

This is the default:
```
fractal.web.theme(
    mandelbrot({
        information: [
            {
                label: 'Built on:',
                value: new Date(),
                type: 'time',
                format: (value) => {
                    return value.toLocaleDateString('en');
                },
            },
        ],
    })
);
```

### DIFF
--- a/packages/mandelbrot/assets/scss/components/_tree.scss
+++ b/packages/mandelbrot/assets/scss/components/_tree.scss
@@ -142,3 +142,8 @@
     padding-top: 0.375rem;
     opacity: 0.7;
 }
+
+.Tree-asideTerm,
+.Tree-asideDescription {
+    display: inline;
+}
--- a/packages/mandelbrot/src/theme.js
+++ b/packages/mandelbrot/src/theme.js
@@ -6,6 +6,53 @@ const Theme = require('@frctl/web').Theme;
 const packageJSON = require('../package.json');
 
 module.exports = function (options) {
+    const labels = _.defaultsDeep(_.clone((options || {}).labels || {}), {
+        info: 'Information',
+        builtOn: 'Built on',
+        search: {
+            label: 'Search',
+            placeholder: 'Search…',
+            clear: 'Clear search',
+        },
+        tree: {
+            collapse: 'Collapse tree',
+        },
+        components: {
+            handle: 'Handle',
+            tags: 'Tags',
+            variants: 'Variants',
+            context: {
+                empty: 'No context defined.',
+            },
+            notes: {
+                empty: 'No notes defined.',
+            },
+            preview: {
+                label: 'Preview',
+                withLayout: 'With layout',
+                componentOnly: 'Component only',
+            },
+            path: 'Filesystem Path',
+            references: 'References',
+            referenced: 'Referenced by',
+            resources: {
+                file: 'File',
+                content: 'Content',
+                previewUnavailable: 'Previews are currently not available for this file type.',
+                url: 'URL',
+                path: 'Filesystem Path',
+                size: 'Size',
+            },
+        },
+        panels: {
+            html: 'HTML',
+            view: 'View',
+            context: 'Context',
+            resources: 'Resources',
+            info: 'Info',
+            notes: 'Notes',
+        },
+    });
     const config = _.defaultsDeep(_.clone(options || {}), {
         skin: 'default',
         rtl: false,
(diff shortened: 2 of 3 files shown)


packages/mandelbrot/views/partials/navigation/information.nunj
```diff
@@ -5,11 +5,21 @@
                 {{ frctl.theme.get('labels.info') }}
             </div>
         </div>
-        <div class="Tree-aside">
-            {{ frctl.theme.get('labels.builtOn') }}:
-            <time datetime="{{ frctl.theme.get('now').toISOString() }}">
-                {{ frctl.theme.get('now').toLocaleDateString(frctl.theme.get('lang')) }}
-            </time>
-        </div>
+        <dl class="Tree-aside">
```
LeBenLeBen, line 8: Shall we add a condition around the wrapper to not display an empty one if no information is provided?

packages/mandelbrot/assets/scss/components/_tree.scss
```diff
@@ -142,3 +142,8 @@
     padding-top: 0.375rem;
     opacity: 0.7;
 }
+
+.Tree-asideTerm,
+.Tree-asideDescription {
+    display: inline;
+}
```
LeBenLeBen, line 149: It happens already with the current code but wondering if we allow more and possibly longer values, maybe we shouldn’t wrap it on multiple line 🤔 

Capture d’écran 2020-10-25 à 19 44 00

packages/mandelbrot/src/theme.js
```diff
@@ -6,6 +6,53 @@ const Theme = require('@frctl/web').Theme;
 const packageJSON = require('../package.json');
 
 module.exports = function (options) {
+    const labels = _.defaultsDeep(_.clone((options || {}).labels || {}), {
```
LeBenLeBen, line 9: Define `options = _.clone(options || {})` on the first line to avoid cloning it twice, once for the `labels` and another for the `config`? Also shouldn't it be `cloneDeep`? Or is cloning required at all? I’m confused 😆

LeBenLeBen: Actually I don’t think I understand why you extracted labels. I have the feeling it does exactly the same as before. What am I missing?

packages/mandelbrot/src/theme.js
```diff
@@ -87,6 +87,24 @@ module.exports = function (options) {
         .filter((url) => url)
         .map((url) => (url === 'default' ? `/${config.static.mount}/css/highlight.css` : url));
 
+    config.information = (
```
LeBenLeBen, line 90: Shouldn’t we explicitly check for `undefined` so users can set `null` or `false` to disable the default in addition to `[]`?

LeBenLeBen (changes requested):
That’s great, much more flexible! We shouldn't forget to update the docs though. Also this makes we want to invest some time to setup e2e tests.

packages/mandelbrot/views/partials/navigation/information.nunj
mihkeleidast: i thought about this, but then it would make more sense for the user to remove information from the navigation panels.

mihkeleidast: Then again, dumb-proofing it so that it wouldn't error out makes sense, too...

packages/mandelbrot/assets/scss/components/_tree.scss
mihkeleidast: Fair, but the current CSS rule affects the entire navigation sidebar, so I'm reluctant to change this here - I think this kind of change should go in the changelog as a separate entry, so maybe a separate issue/pr would be better?

packages/mandelbrot/src/theme.js
mihkeleidast: Ah yes, I changed where I assigned the default information entry at the last minute, so now this change is not necessary at all. Good catch 👍

packages/mandelbrot/src/theme.js
mihkeleidast: Nah I don't think so, the most correct (also perf friendly) way to disable the information section would be to remove it from `config.nav`. This is also similar to how other options behave, config.nav and config.panels, for example.

mihkeleidast:
I'll file a docs issue and maybe you want to file an issue for mandelbrot e2e tests? I wouldn't want to start investigating e2e tests for this PR.

packages/mandelbrot/assets/scss/components/_tree.scss
LeBenLeBen: Agree

LeBenLeBen (approved):

PR #684, opened 2020-11-07
docs: add Nunjucks example

LeBenLeBen:
Here is a new example app, this time using the Nunjucks adapter. The goal here isn’t to run some integration or e2e tests like for the Handlebars example (even if it could still be added later though) but rather to showcase what Fractal offers.

I tried to create common components and examples combining them together with `render`, `extends` and context inheritance.

I would like to deploy this app somewhere, for example on demo.fractal.build; probably hosted on Netlify (I would need help or access for that). After that, I’ll submit other PRs to update the old screenshots we have here and there and link to this demo where it makes sense.

PS: I went with Webpack to bundle the CSS, firstly to stay consistent with Mandelbrot and avoid new dependencies and secondly to push even further the example, using one of the most popular tool out there. This way it could be extended later to support JavaScript or other stuff. However, it’s a bit over-the-top and could be replaced by just Postcss, with likely better performances.

My apologies for the very large diff 😬 

Screenshot_2020-11-07 Home Fractal Nunjucks example


mihkeleidast:
I'm not convinced this should be part of the main repo if it does not provide any test coverage value. I also would not want to maintain this if tailwind (or something else used by this example) has updates or security issues etc.

Keeping it in a separate repo would also help with the netlify build/deploy, I think, plus it would be more convenient to use it as a starter (just fork the repo, vs how do you fork as single package from this monorepo?).

examples/nunjucks/assets/css/main.css
```diff
@@ -0,0 +1,4 @@
+/* stylelint-disable scss/at-rule-no-unknown */
```
mihkeleidast, line 1: can't we define these "unknown" at rules in stylelint config or somewhere else, so there would not be a need for the ignore?

examples/nunjucks/components/01-atoms/button/button--link.njk
```diff
@@ -0,0 +1,5 @@
+{% extends '@base-button' %}
+
+{% block class %}text-pink-600 hover:text-pink-500 focus:text-pink-500 font-bold uppercase tracking-wide transition-all duration-200 ease-out group{% if class %} {{ class }}{% endif %}{% endblock class %}
+
+{% block after %}{% render '@icon--tiny', { id: 'chevron-right', class: 'ml-2 transition-all duration-200 ease-out transform group-hover:translate-x-1' }, true %}{% endblock after %}
```
mihkeleidast, line 5: i guess a personal preference, but i really really dislike when so much code is on a single line. it's partially inevitable if using tailwind but we could at least put object properties on a newline?

examples/nunjucks/components/_preview.njk
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html class="antialiased text-gray-800">
+    <head>
+        <meta charset="utf-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1">
+        <meta http-equiv="X-UA-Compatible" content="IE=edge">
+
+        <title>Preview Layout</title>
```
mihkeleidast, line 8: value of title could be dynamic, e.g. we can use _self.title (or something) here

examples/nunjucks/assets/css/main.css
LeBenLeBen: Unlike EsLint, Stylelint doesn’t seem to accept an `overrides` section in its config, allowing to change rules for specific files, or at least I didn’t find something in that direction in the docs. Now I’m thinking it might accept sub-config files per folder, but it doesn’t really matter if we move that to another repo.

examples/nunjucks/components/01-atoms/button/button--link.njk
LeBenLeBen: Agree, I forgot I could use the `{%- -%}` syntax to keep the template cleaner while removing white-spaces, should be better now.

examples/nunjucks/components/_preview.njk
LeBenLeBen: I always thought this was useless, but it actually isn’t! It’s obviously the title of the preview page, how could I have missed that. Thanks for the hint, it’s fixed.

LeBenLeBen:
@mihkeleidast At first I thought it would be useful to be close to the code base to test while developing, but I must admit I don’t really see the added value of so many components; a simpler example app with just ugly bare components like we have for Handlebars would do the job just as well. 

And I also think it would be difficult to keep that working for both a demo purpose and automated testing purpose, pretty sure these two use cases will conflicts at some point.

Let’s make this a separate repo then, can you create me one like `demo.fractal.build` or whatever name you prefer?

LeBenLeBen:
Moving to https://github.com/frctl/demo.fractal.build

examples/nunjucks/assets/css/main.css
mihkeleidast: Not overrides per se, but you should be able to define at-rules that should not throw: https://github.com/kristerkari/stylelint-scss/blob/HEAD/src/rules/at-rule-no-unknown/README.md#ignoreatrules-regex-string

PR #685, opened 2020-11-08
feat(Mandelbrot): display preview iframe size

### LINKED ISSUE #606: Add configurable preview window sizes/switcher

LeBenLeBen:
Relates to #606


packages/mandelbrot/assets/js/components/preview.js
```diff
@@ -60,6 +62,28 @@ export default class Preview {
         });
     }
 
+    _updateIframeRefs() {
+        this._iframe = this._el.find('[data-role="window"]');
+        this._previewIframeWindow = this._iframe ? this._iframe.get(0).contentWindow : null;
+        this._previewSize = $('[data-role="preview-size"]');
+
+        if (this._iframe) {
+            this._iframe.on('load', this._updateSize.bind(this));
```
mihkeleidast, line 71: `.bind(this)` always creates a new function. since you do it twice here, it would be better to cache the bound method in a variable. maybe even in the constructor, since this method itself is called on an event and we don't know how many times the event will be called.

mihkeleidast (changes requested):
image

Should we maybe align the measurements to right? Currently it's "centered between the title and the status tag", which means it'll start jumping a bit when the title of the component varies. It should move a bit less when aligned to right because while the statuses can differ, they are still predefined and a bit shorter in text.

Regarding closing #606, please don't close it with this PR. While I agree with your sentiment about browser tools providing all you need as a developer, not everyone who uses Fractal are devs :) So I still see value in supporting predefined size switcher.

packages/mandelbrot/assets/js/components/preview.js
LeBenLeBen: 👍

packages/mandelbrot/assets/js/components/preview.js
```diff
@@ -60,6 +65,28 @@ export default class Preview {
         });
     }
 
+    _updateIframeRefs() {
+        this._iframe = this._el.find('[data-role="window"]');
+        this._previewIframeWindow = this._iframe ? this._iframe.get(0).contentWindow : null;
+        this._previewSize = $('[data-role="preview-size"]');
```
mihkeleidast, line 71: Not sure about the global query here - I think everywhere else element queries are scoped to `this._el`. Maybe it would be better to query this element in Pen class and pass it into the Preview constructor?

LeBenLeBen: Like this?

mihkeleidast: Yup, looks good now 👍

mihkeleidast (approved):

PR #713, opened 2020-12-21
refactor: use get-port instead of portscanner

mihkeleidast:
My initial goal was to remove the use of `Promise.promisfyAll` but when I looked at the status of the dependency it seemed like a good idea to replace it altogether. [portscanner](https://github.com/baalexander/node-portscanner) has not been updated in 2 years and it has functionality we don't use. [get-port](https://github.com/sindresorhus/get-port) has only the functionality we need and it's maintained by Sindre Sorhus, so I would expect it to remain updated in the future. [According to npm trends](https://www.npmtrends.com/portscanner-vs-get-port), get-port is also about 10x smaller in size and has no additional dependencies, while portscanner has two.


LeBenLeBen (commented):
If you start the Handlebars example (which use sync mode) twice, you get an error saying 3001 is already in use. Is that expected behavior? That would prevent two separate Fractal instance with sync from running at the same time. I would expect this code to be able to find an alternate port instead of crashing?!

Edit: no issue when not using sync mode.

mihkeleidast:
Hm, that is working from me. I ran `npx lerna exec --scope=@frctl/handlebars-example -- npm run start` from root in two separate terminals. The second server starts on port 3003.

There is a [warning about that in the get-port README](https://github.com/sindresorhus/get-port#beware), but AFAIK that race condition happened with portscanner as well (it has happened to me before).

LeBenLeBen:
I tried with your command but the result is the same. I opened the first server in the browser and waited a bit but I was never able to start another one in parallel. Before I start the second one, I can properly see the ports being busy:

```
node       7502  leben   24u  IPv6 0x8541de7140103ab1      0t0  TCP *:3001 (LISTEN)
node       7502  leben   25u  IPv6 0x8541de71401040d1      0t0  TCP *:3000 (LISTEN)
node       7502  leben   26u  IPv6 0x8541de7140102e71      0t0  TCP *:3002 (LISTEN)
```

The problem seems to occur only when it’s another Fractal instance that occupy the port already, not with a completely different app. I’ve seen issues of people reporting similar problems and saying it works with https://github.com/http-party/node-portfinder which seems slightly more popular. Maybe I can try to set it up and see if that fixes the issue?

mihkeleidast:
Well, that's weird... I succesfully started many parallel servers and no issues whatsoever. We can test node-portfinder but frankly I don't like their Promise API very much - it's weird to access a Promise version from a different submethod. Maybe I'll revert this to use portscanner, if that does not cause issues for you? I can only remove the promisification since the module exposes a reasonable Promise API.

LeBenLeBen (approved):
It seems that this problem was still occurring before, so I guess that’s fine to go with it and improve later if others encounter the same behavior.

PR #774, opened 2020-12-25
feat: style native select instead of using a plugin

mihkeleidast:
Select2 is only used in a single place (assets file browser) in Mandelbrot, but accounts for ~30% of the JS bundle (did not analyze the CSS size, but I would presume the savings are also big). I know Mark had some additional ideas on how to use the Select2 dependency better but those never came to fruition.

### DIFF
--- a/packages/mandelbrot/assets/scss/components/_file-browser.scss
+++ b/packages/mandelbrot/assets/scss/components/_file-browser.scss
@@ -22,18 +22,15 @@
     position: relative;
     appearance: none;
     min-width: 15em;
-    padding: 0 24px 0 8px;
-    border: 1px solid rgba(83, 83, 99, 0.25);
-    border-radius: 4px;
-    font-size: 0.875rem;
-    color: #444;
-    line-height: 28px;
-    height: 28px;
+    padding: 0.3125rem 2rem 0.3125rem 0.5625rem;
+    border: 1px solid $color-frame-border;
+    border-radius: 3px;
+    height: 36px;
     cursor: pointer;
     background-color: #fff;
     background-image: url("data:image/svg+xml;charset=US-ASCII,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20width%3D%22292.4%22%20height%3D%22292.4%22%3E%3Cpath%20fill%3D%22%23888888%22%20d%3D%22M287%2069.4a17.6%2017.6%200%200%200-13-5.4H18.4c-5%200-9.3%201.8-12.9%205.4A17.6%2017.6%200%200%200%200%2082.2c0%205%201.8%209.3%205.4%2012.9l128%20127.9c3.6%203.6%207.8%205.4%2012.8%205.4s9.2-1.8%2012.8-5.4L287%2095c3.5-3.5%205.4-7.8%205.4-12.8%200-5-1.9-9.2-5.5-12.8z%22%2F%3E%3C%2Fsvg%3E");
     background-repeat: no-repeat;
-    background-position: right 0.5em top 55%;
+    background-position: right 0.65em top 52%;
     background-size: 0.65em auto;
 
     &::-ms-expand {
@@ -43,11 +40,6 @@
     &::-ms-value {
         color: inherit;
         background: inherit;
-    }
-
-    &:focus {
-        outline: none;
-        border-color: #000;
     }
 }
 
--- a/packages/mandelbrot/assets/scss/components/_file-browser.scss
+++ b/packages/mandelbrot/assets/scss/components/_file-browser.scss
@@ -7,19 +7,6 @@
     margin: -0.75rem -0.75rem 0.75rem -0.75rem;
     padding: 0.5rem 0.75rem;
     background-color: rgba($color-frame-border, 0.05);
-
-    .select2-container {
-        @include font(caption);
-        @include margin-inline(start, -0.25rem);
-
-        margin-bottom: -1px;
-        min-width: 15em;
-    }
-
-    select:focus,
-    select:active {
-        outline: none;
-    }
 }
 
 .FileBrowser-select-label {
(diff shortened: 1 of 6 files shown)


packages/mandelbrot/assets/scss/components/_file-browser.scss
```diff
@@ -31,36 +18,37 @@
...
+    border-radius: 4px;
+    font-size: 0.875rem;
+    color: #444;
+    line-height: 28px;
+    height: 28px;
+    cursor: pointer;
+    background-color: #fff;
+    background-image: url("data:image/svg+xml;charset=US-ASCII,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20width%3D%22292.4%22%20height%3D%22292.4%22%3E%3Cpath%20fill%3D%22%23888888%22%20d%3D%22M287%2069.4a17.6%2017.6%200%200%200-13-5.4H18.4c-5%200-9.3%201.8-12.9%205.4A17.6%2017.6%200%200%200%200%2082.2c0%205%201.8%209.3%205.4%2012.9l128%20127.9c3.6%203.6%207.8%205.4%2012.8%205.4s9.2-1.8%2012.8-5.4L287%2095c3.5-3.5%205.4-7.8%205.4-12.8%200-5-1.9-9.2-5.5-12.8z%22%2F%3E%3C%2Fsvg%3E");
+    background-repeat: no-repeat;
+    background-position: right 0.5em top 55%;
```
LeBenLeBen, line 36: 52% is more visually centered vertically to me with a 36px height

packages/mandelbrot/assets/scss/components/_file-browser.scss
```diff
@@ -31,36 +18,37 @@
...
     min-width: 15em;
-
-    &[aria-selected="true"] {
-        background-color: rgba($color-frame-border, 0.1) !important;
+    padding: 0 24px 0 8px;
+    border: 1px solid rgba(83, 83, 99, 0.25);
+    border-radius: 4px;
+    font-size: 0.875rem;
+    color: #444;
+    line-height: 28px;
```
LeBenLeBen, line 30: It‘s not perfectly centered for me on Firefox, seems that `1` without unit have a better result in all browsers I have tested. Or actually no value at all, I think browsers vertically center text in native select by default.

packages/mandelbrot/assets/scss/components/_file-browser.scss
```diff
@@ -31,36 +18,37 @@
...
-        outline: none;
-        border-color: $color-link;
-        border-color: var(--border-color, $color-link);
-    }
-}
-
-.select2-results__option {
-    @include font(caption);
-
+.FileBrowser-select {
```
LeBenLeBen, line 21: I’m wondering if we should extract it as a global `.Select` that we could easily reuse later on if necessary. Would be great it shares as many properties as possible with the search input (colors, radius, focus state, …)

packages/mandelbrot/assets/scss/components/_file-browser.scss
```diff
@@ -31,36 +18,37 @@
...
+        color: inherit;
+        background: inherit;
+    }
 
-    background-color: $color-header-background !important;
-    background-color: var(--background-color, $color-header-background) !important;
-    color: $color-header-content !important;
-    color: var(--color, $color-header-content) !important;
+    &:focus {
+        outline: none;
```
LeBenLeBen, line 49: I wouldn’t suppress the browser default without adding a large box-shadow instead, border isn’t enough to me to be easily found in the page. To be sync with the search field as well I think.

packages/mandelbrot/assets/scss/components/_file-browser.scss
```diff
@@ -31,36 +18,37 @@
...
-    @include font(caption);
-
+.FileBrowser-select {
+    position: relative;
+    appearance: none;
     min-width: 15em;
-
-    &[aria-selected="true"] {
-        background-color: rgba($color-frame-border, 0.1) !important;
+    padding: 0 24px 0 8px;
```
LeBenLeBen, line 25: If you agree on making it higher, I would increase the side padding by 2px

LeBenLeBen (changes requested):
It’s been a long time I was bothered by this dep being barely used, that‘s great improvement 👍

packages/mandelbrot/assets/scss/components/_file-browser.scss
```diff
@@ -31,36 +18,37 @@
...
+    border: 1px solid rgba(83, 83, 99, 0.25);
+    border-radius: 4px;
+    font-size: 0.875rem;
+    color: #444;
+    line-height: 28px;
+    height: 28px;
```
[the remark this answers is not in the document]
mihkeleidast: Oh it totally makes sense to align the styles with the search input 👍

packages/mandelbrot/assets/scss/components/_file-browser.scss
mihkeleidast: Removing it completely did not seem to affect it at all 👍

packages/mandelbrot/assets/scss/components/_file-browser.scss
mihkeleidast: I wouldn't oppose that, though looking at how the select label is styled currently, it's highly specific to the file browser. I would leave it as it is right now. We can refactor it out when another usecase pops up in the future.

packages/mandelbrot/assets/scss/components/_file-browser.scss
mihkeleidast: Gave it the exact same paddings as the Search input has, it should be fine like this.

packages/mandelbrot/assets/scss/components/_file-browser.scss
mihkeleidast: I removed the custom focus styles completely. I played around with custom because I happened to test in IE11 and there were no focus styles visible at all... then again, the search input also does not have focus styles in IE11. Maybe it's not a too bad to leave it a bit buggy in IE?

packages/mandelbrot/assets/scss/components/_file-browser.scss
```diff
@@ -31,36 +18,37 @@
...
+        background: inherit;
+    }
 
-    background-color: $color-header-background !important;
-    background-color: var(--background-color, $color-header-background) !important;
-    color: $color-header-content !important;
-    color: var(--color, $color-header-content) !important;
+    &:focus {
+        outline: none;
+        border-color: #000;
```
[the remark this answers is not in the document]
mihkeleidast: Removed the custom focus style completely.

packages/mandelbrot/assets/scss/components/_file-browser.scss
LeBenLeBen: 👍

packages/mandelbrot/assets/scss/components/_file-browser.scss
LeBenLeBen: Yeah no need to workaround that for a browser that will be considered dead in 6 months.

LeBenLeBen (approved):

PR #779, opened 2020-12-28
test: add twig example with integration tests

mihkeleidast:
Basically, this is copy-paste from the handlebars example, with Twig-specific changes - removed some tests for helpers that the Twig adapter does not have, added some for Twig-specific things. Twig adapter still does not have a full coverage, however - there's some code that is dependent on the adapter options when registering it and I wasn't too sure how to structure that logic in the example package. Maybe in a future PR :)

The tests and snapshots are mostly 100% same. Is there a reasonable way how we could share the tests so they would not be duplicated & would ensure adapters are implemented similarly?


examples/twig/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+to start Fractal server in development mode, or:
+
+```
+npm run build
+```
+
+to build a static version in the `dist` folder.
+
+## Tests
```
LeBenLeBen, line 21: Now that tests are defined in the root package.json, this section should either be removed from this file (and Handlebars example) or rephrased to specify this must be run from the root directory.

LeBenLeBen (changes requested):
I’m wondering if it wouldn’t be better to keep as many tests as possible shared between examples? Wouldn’t that be easier to maintain if the behavior is supposed to be the same?

mihkeleidast:
> I’m wondering if it wouldn’t be better to keep as many tests as possible shared between examples? Wouldn’t that be easier to maintain if the behavior is supposed to be the same?

What do you think the most reasonable way for sharing the tests would be?

examples/twig/README.md
mihkeleidast: Yup, removed from both.

mihkeleidast:
I created another package to hold the shared adapter tests - not sure if this is the best way to do this but anyhow this should be enough for now.

The gist should be that if an adapter implements a feature, it should:
- pass all the common tests
- pass any additional tests if the adapter has any specific additional functionality

The Handlebars adapter injects some weird additional newlines with the render tag so the tests are weird in that place and the twig templates have additional unnecessary newlines.

LeBenLeBen (approved):
💯

PR #808, opened 2021-01-14
test: refactor markdown renderer tests to jest

mihkeleidast:


packages/core/src/markdown.js
```diff
@@ -27,6 +27,7 @@ module.exports = function markdown(content, mdConfig) {
     return marked(_.toString(content), mdConfig);
 };
 
+// TODO: remove if noone understands what this is for
```
mihkeleidast, line 30: Don't know about this, would've added a test but I have no idea what this is supposed to do. It is only used in DocSource in a method named `toc` but that is where the breadcrumbs end - nowhere is this used internally.

LeBenLeBen: I’ve looked the commit (c0924a245d1a01c414c118f9fd2cb94a9d5fe931) introducing this feature, it seems to add `id="slug"` to markdown headings. However I don’t get how this is all wired together.

LeBenLeBen (approved):

packages/core/src/markdown.js
mihkeleidast: Yeah I don't either, let's remove when doing a major version at some point & add it back later if someone depended on it :)

PR #809, opened 2021-01-14
test: refactor core utils test to jest

mihkeleidast:


packages/core/test/utils.spec.js
```diff
@@ -0,0 +1,309 @@
...
+                mode: 'django',
+                scope: 'text.html.django',
+                color: null,
+            };
+            expect(utils.lang('file.nunjucks')).toEqual(expected);
+            expect(utils.lang('file.nunjs')).toEqual(expected);
+            expect(utils.lang('file.nunj')).toEqual(expected);
+            expect(utils.lang('file.nj')).toEqual(expected);
+            expect(utils.lang('file.jinja2')).toEqual(expected);
+            expect(utils.lang('file.j2')).toEqual(expected);
```
LeBenLeBen, line 27: The [official extension](https://mozilla.github.io/nunjucks/templating.html) is `njk`, you skipped it because you expect `fang` to cover that already?

packages/core/test/utils.spec.js
```diff
@@ -0,0 +1,309 @@
...
+
+    describe('.titlize()', () => {
+        it('returns titlized string', () => {
+            expect(utils.titlize('not-a-title')).toEqual('Not A Title');
+        });
+    });
+
+    describe('.slugify()', () => {
+        it('returns slugified string', () => {
+            expect(utils.slugify('Not A Slug')).toEqual('not-a-slug');
```
LeBenLeBen, line 49: Maybe worth adding a test with some special char that could break an URL like `?`

packages/core/test/utils.spec.js
```diff
@@ -0,0 +1,309 @@
...
+
+            mock.restore();
+        });
+    });
+
+    describe('.md5()', () => {
+        it('returns same hash for same string', () => {
+            const first = utils.md5('string');
+            const second = utils.md5('string');
+            expect(first).toEqual(second);
```
LeBenLeBen, line 131: Worth checking it’s a string of some length or at least not empty/null/undefined?

packages/core/test/utils.spec.js
mihkeleidast: yup, we don't have a switch case in our code for that extension.

packages/core/test/utils.spec.js
mihkeleidast: Hmm, actually the function does not do anything with special characters like that. 

It does, however, replace some unique letters with the more common ones, like `expect(utils.slugify('slug-šäöüõ-slug')).toEqual('slug-saouo-slug');`. so I've added a test for that.

packages/core/test/utils.spec.js
mihkeleidast: Added an actual md5 value assertion.

LeBenLeBen (approved):

PR #825, opened 2021-01-21
fix: skip linkRefs if there are no references

mjomble:
If `item.references` is an empty array, an empty regex is created, which matches all empty strings between any two consecutive characters.

So if the input string contains `n` characters, the handle callback is invoked `n + 1` times.
As the parameter is an empty string, an error is thrown and caught each time.
I discovered this in a project where it added several seconds to the page render time.

This PR fixes the issue.

### DIFF
--- a/packages/mandelbrot/src/filters.js
+++ b/packages/mandelbrot/src/filters.js
@@ -65,7 +65,7 @@ module.exports = function (theme, env, app) {
     });
 
     env.engine.addFilter('linkRefs', function (str, item) {
-        if (!(item.isComponent || item.isVariant)) {
+        if (!(item.isComponent || item.isVariant) || !item.references.length) {
             return str;
         }
         const refs = item.references;
--- a/packages/mandelbrot/src/filters.js
+++ b/packages/mandelbrot/src/filters.js
@@ -65,7 +65,10 @@ module.exports = function (theme, env, app) {
     });
 
     env.engine.addFilter('linkRefs', function (str, item) {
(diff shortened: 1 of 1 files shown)


packages/mandelbrot/src/filters.js
```diff
@@ -65,7 +65,7 @@ module.exports = function (theme, env, app) {
     });
 
     env.engine.addFilter('linkRefs', function (str, item) {
-        if (!(item.isComponent || item.isVariant)) {
+        if (!(item.isComponent || item.isVariant) || !item.references.length) {
```
mihkeleidast, line 68: Not sure what other types of items can be passed in here besides components/variants, but if `item.isComponent === false` and `item.isVariant === false`, then if `item.references === undefined`, this will error out. Maybe do a second if with an early return instead?

Or... maybe I'm reading this wrong because the condition is a bit complex :( I think even if my previous explanation is not true, a second if clause with an early return would improve readability.

mihkeleidast (changes requested):

packages/mandelbrot/src/filters.js
mjomble: Done. It's functionally equivalent to the previous commit, but indeed more readable 🙂

mihkeleidast (approved):

packages/mandelbrot/src/filters.js
mihkeleidast: Conditions with multiple or's and negations make my brain hurt :) Much better now.

PR #831, opened 2021-01-24
chore(examples): use correct syntax to customize theme

LeBenLeBen:
The current syntax is wrong and doesn’t work. I also removed the panels list since it was the same as the default. Added a link to the available options instead. I know this basically does nothing but it’s useful when you want to quickly tweaks some Mandelbrot settings while working on or reviewing changes.


mihkeleidast (approved):
Ack, I've noticed this as well but never bothered to change it :)

PR #842, opened 2021-01-31
perf: ignore excluded directories from file watcher

mihkeleidast:
We already skip parsing the file system for excluded directories, but they are still watched by chokidar. This means that in my actual project with a lot of excluded directories (monorepo with a bunch of node_modules directories), the chokidar file watcher takes a lot of time to start (and uses A LOT of cpu), because there are simply a lot of directories to watch.

Some stats:
```
BEFORE:
watching 15167 directories
watch start time: 51623.770ms


AFTER:
watching 525 directories
watch start time: 903.210ms
```

So I guess it's obvious how much faster Fractal startup is going to be after this change :)


packages/core/src/mixins/source.js
```diff
@@ -112,8 +112,9 @@ module.exports = mixin(
                 const sourcePath = this.fullPath;
                 if (!this._monitor && sourcePath) {
                     Log.debug(`Watching ${this.name} directory - ${sourcePath}`);
+                    const exclude = this.get('exclude') || [];
```
LeBenLeBen, line 115: `get` has a second argument for the default value but somehow this isn’t used much anywhere…

LeBenLeBen (approved):
Great improvement 👏

packages/core/src/mixins/source.js
mihkeleidast: As `get` is not typed, specifying the default value outside with or operator gives code editors a bit more knowledge about the variable, I think.

PR #873, opened 2021-02-10
test: refactor core collection mixin tests to jest

mihkeleidast:


packages/core/src/entities/collection.js
```diff
@@ -23,10 +23,6 @@ module.exports = class EntityCollection extends mix(Heritable, Collection, Entit
         this.setProps(config);
     }
 
-    get isRoot() {
```
mihkeleidast, line 26: This was just weird because the collection mixin did not set isRoot property on it's own so the `.rootCollections()` method on the mixin could not work without instantiating entities. The change makes sense for me & did not break the root collections functionality in my testing.

LeBenLeBen (approved):

PR #878, opened 2021-02-11
test: add tests for heritable mixin

mihkeleidast:
Added some type doc as well because I'm pretty tired of seeing "any" everywhere in my editor when it may be derived from types. It may also help if we want to add TypeScript parsing/validating at some point.


packages/core/src/mixins/heritable.js
```diff
@@ -38,10 +54,14 @@ module.exports = mixin(
...
             }
 
+            /**
+             * @return {string[]}
+             */
             getHeritable() {
                 if (this._heritable) {
                     return Array.from(this._heritable);
                 }
+                // TODO: how to get here?
```
mihkeleidast, line 64: So my logic goes:
- `this._parent` can only be set via `.setHeritable()`
- if `this._parent` is set via `.setheritable()`, it will also set `this._heritable`, so it won't arrive here
- the only way to set `_parent` without `_heritable` is to set it directly, but that should not be allowed since `_parent` is a private property.

I guess I'd leave it as it is for now, maybe in the future when we have more test coverage we can remove it.

LeBenLeBen (approved):

PR #884, opened 2021-02-12
fix(handlebars): fix render tag resolving escaped component handles to context object

### LINKED ISSUE #879: Nested render helpers inside partial-block
I have a Fractal instance that uses nested `render` helpers inside a `@partial-block`. This method worked in `@frctl/handlebars` version 1.1.5 but fails when I upgrade that package.

1. Clone the reduced test case repo below, run `npm install`, then `npm start`
2. Visit http://localhost:3000/components/detail/carousel-card-grid and observe that the Card Grid Carousel loads as expected (carousel has 2 slides of 2 cards each)
3. Stop the dev server and update the `@frctl/handlebars` version in `package.json` to the latest (currently 1.2.6)
4. Run `npm install` again then `npm start`
5. Visit http://localhost:3000/components/detail/carousel-card-grid again and observe that the cards do not load in the carousel

**Reproduces how often:**

100%

### Reduced test case

https://github.com/ccardinaux/fractal-dev-sandbox

### Context

-   Fractal version: 1.5.3
-   `@frctl/handlebars` version: 1.2.6
-   Node version: 12.16.2
-   OS: MacOS 10.15.7

mihkeleidast:
Fixes #879.

- It basically resolved the context twice - since the `defaultContext` was already resolved context object it should not be necessary to resolve it again.
- Twig adapter already behaved correctly.
- This probably should be changed in Nunjucks adapter as well but I don't want to change it without having integration tests in place.


LeBenLeBen (approved):
I think it’s a bit difficult to understand what is expected exactly in the test because the description is negative and the assertion positive, I don’t know if we can phrase it better. Otherwise it looks good!

mihkeleidast:
Yeah thought about it a bit but I don't see how this can be worded better. In Fractal, "resolving handles in context to context objects" is the default behavior so testing against that "when handle is escaped" makes sense to me. But I get your point about positive/negative test/assertions and I'll try to do better in the future :D

PR #885, opened 2021-02-13
test: add nunjucks adapter example with integration tests

mihkeleidast:
Copy-paste of the Handlebars/Twig example. Do not full coverage on the extensions because no existing tests had been written for them and the equivalent helpers in the Handlebars adapter are not covered as well. I would rather add tests for them together after this is merged.


packages/nunjucks/README.md
```diff
@@ -20,7 +20,7 @@ npm i @frctl/nunjucks --save
 
 ```javascript
 fractal.components.engine("@frctl/nunjucks"); // register the Nunjucks adapter for your components
-fractal.components.set("ext", ".nunj"); // look for files with a .nunj file extension
+fractal.components.set("ext", ".njk"); // look for files with a .nunj file extension
```
mihkeleidast, line 23: https://mozilla.github.io/nunjucks/templating.html#file-extensions

njk seems to be the community-adopted default so I think we should recommend it as well?

Issue #605, opened 2016-08-08
Code highlighting on large files slows down page load significantly


allmarkedup:
Need to handle this differently (or at least improve caching at a bare minimum)

mihkeleidast:
I did some testing for #906 and there does not seem to be a remarkable win when caching the highlight result. Maybe this was a bigger issue 5 years ago but I think currently the page load times are bottlenecked by other things.

PR #906, opened 2021-02-19
perf: cache highlighted code

mihkeleidast:
See #605.

I POCed this out to see how much faster page loads / build this will get us, but...
- on a large project, total build time was reduced approx 10s (that would be approximately 2x reduction in time spent highlighting code). not a huge win to be honest, since that project builds for many minutes.
- in server mode, I did not see any very obvious wins. I looked at some pages that had components over 1000 lines long, etc, but not too much changed (looked at TTFB) without the cache vs with.

So I could probably also not merge this, but filing anyway to see what others think. Maybe I'm just bad at testing this stuff :)

P.S. I _think_ the code beautifier in mandelbrot might also take a lot of time but I haven't tested that out. Maybe adding a cache there would also bring some benefits?

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "liftoff": "^3.1.0",
     "lodash": "^4.17.20",
     "log-update": "^4.0.0",
+    "lru-cache": "^6.0.0",
     "marked": "^2.0.0",
     "mime": "^2.5.2",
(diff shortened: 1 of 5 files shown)


mihkeleidast:
I'll bring in the tests in a separate new PR but I don't think this is worth it otherwise.

PR #940, opened 2021-03-23
fix: skip nunjucks path filter, when path is not defined

thomasaull:
Using nunjucks with the following `context.js`:

```js
module.exports = {
  context: {
    default: {
      source: '/videos/my-video.mp4',
      poster: null,
    },
  },
};
```
and doing this in the `.nunjucks` will throw an error on build, because the `path` parameter is undefined:
```
{% set default = objectAppend(default, 'poster', default.poster | path) %}
```
Is it reasonable to just skip the path filter when the path is not defined? This way we can avoid a lot of additional logic in our templates.

### DIFF
--- a/packages/nunjucks/src/filters/path.js
+++ b/packages/nunjucks/src/filters/path.js
@@ -5,6 +5,8 @@ const utils = require('@frctl/core').utils;
 
 module.exports = function (fractal) {
     return function (path) {
+        if (!path) return path;
+
         let env = this.lookup('_env');
         let request = env.request || this.lookup('_request') || this.ctx.request;
 


mihkeleidast:
I'm personally more of a "don't call a function if required params are missing" guy instead of doing early returns in functions. But it might be safe to do this in a library setting, especially in this kind of templating, where there is no type checking other than getting runtime errors :)

mihkeleidast (changes requested):
I would ask you to:
- add the same behavior to all the adapters (this is just as likely to come up with other templating engines)
- maybe add a integration test so this would be covered? "does not throw when path argument is missing" or something like that would be good.

mihkeleidast:
Note that #886 and #885 should probably be merged first to add tests...

thomasaull:
@mihkeleidast Fair enough 😅

PR #1004, opened 2021-04-29
chore: drop node 10 support

mihkeleidast:
Node 10 will be EOL tomorrow! Bumped the release script to use Node 14, since that's the current Active LTS.


Chapabu:
Closing as this is waaaay out of date and fixed with Node 14 in #1206

PR #1061, opened 2021-07-13
URL encode `:` to `%3A` in pattern urls

13twelve:
Stops `Failed to launch 'foo:btn' because the scheme does not have a registered handler.` type errors on Fractal build generated static navigation.

If your component config file specifies a handle like: `"handle": "foo:btn",` or if you use a pattern with a `:` in the name like "ratio-box--16:9" then all will be fine for local development but after a Fractal build, the statics will not navigate.

My primary use case is name spacing of components. I have one Fractal instance to demonstrate patterns of 2 web sites served from a multisite Craft app. I have website Foo and website Bar, they share a few components but have site specific components also - the kicker being, some of them have the same name. And so I need to be able to use the correct version of a component within other components and on final templates.

Eg: 

```twig
{% include '@foo:btn' with data %} // outputs `btn` from `foo`
{% include '@bar:btn' with data %} // outputs `btn` from `bar`
```

### DIFF
--- a/packages/mandelbrot/src/filters.js
+++ b/packages/mandelbrot/src/filters.js
@@ -10,13 +10,13 @@ module.exports = function (theme, env, app) {
             if (!item.path) {
                 return '/';
             }
-            return theme.urlFromRoute('page', { path: item.path });
+            return theme.urlFromRoute('page', { path: item.path }).replace(/:/g, '%3A');
         } else if (item.isComponent || item.isVariant) {
-            return theme.urlFromRoute('component', { handle: item.handle });
+            return theme.urlFromRoute('component', { handle: item.handle }).replace(/:/g, '%3A');
         } else if (item.isAssetSource) {
-            return theme.urlFromRoute('asset-source', { name: item.name });
+            return theme.urlFromRoute('asset-source', { name: item.name }).replace(/:/g, '%3A');
         } else if (item.isAsset) {
-            return Path.join('/', app.get('web.assets.mount'), item.srcPath);
+            return Path.join('/', app.get('web.assets.mount'), item.srcPath).replace(/:/g, '%3A');
         }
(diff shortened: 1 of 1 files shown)


packages/mandelbrot/src/filters.js
```diff
@@ -10,13 +10,13 @@ module.exports = function (theme, env, app) {
             if (!item.path) {
                 return '/';
             }
-            return theme.urlFromRoute('page', { path: item.path });
+            return theme.urlFromRoute('page', { path: item.path }).replace(/:/g, '%3A');
```
mihkeleidast, line 13: The `theme.urlFromRoute` function is used elsewhere as well. Would it maybe make more sense to add that replacing inside that function? There's already a cleaning function there that may be the proper place: https://github.com/frctl/fractal/blob/main/packages/web/src/theme.js#L147-L149

mihkeleidast (commented):
Can you also add a test for the fix?

packages/mandelbrot/src/filters.js
13twelve: I tried this but `urlFromRoute` is also used when generating file names and so produces HTML files with `%3A` in their names. I'm not discounting a better place for this `replace` though, I'm just not sure where would be better.

mihkeleidast: By "elsewhere" I did not mean other themes, rather that it's used directly in some nunjucks templates as well. If that was to be added everywhere, that's a lot of duplication TBH.

I can take a look sometime, maybe I can come up with some ideas on how to bypass the filename issue.

13twelve: Ack on further testing this doesn't work while the server is running, it fouls up the ajax page navigation. Going to rethink!

13twelve:
I'm going to try and find a way to only make the replace happen for static builds

PR #1062, opened 2021-07-14
add relative path to component links in static builds

13twelve:
### Purpose of update:

Having a `:` in your component handle would produce a static URL like:`href="foo:btn.html"`
The browser tries to interpret this as an app link, which gives you a `Failed to launch 'foo:btn' because the scheme does not have a registered handler` error message in the browser developer console and stops page navigation.

Having a relative path current folder `./`, so producing `href="./foo:btn.html"` stops the browser interpreting as an app link.

For server builds, this doesn't apply, the existing paths are used as before.

I can't lie, changing paths in the build of a project I'm not super familiar with gives me the fear. The tests pass, although I'm not sure anything tests following links in static builds. 

### What are app links?

Firstly, apologies if you're already familiar, I wasn't, or at least, I wasn't aware that beyond `tel://` they are a thing. To open the Twitter app from a link if you have it installed `<a href="twitter://user?screen_name=opticalcortex">open</a>`. 

It seems the browser just needs the `:` and not the `://` to interpret a link as an app link. And so using `:` in a Fractal handle, the browser fails to find an appropriate app and so emits an error message in the console and stops.

### Why use `:` in a component handle?

Components may naturally have `:` in their name, such as `ratio-box--16:9` 
or a user maybe name spacing components in a multisite Fractal library, 
where components may have the same name but live in different sites:

```
{% include '@foo:btn' with data %} // outputs `btn` from site `foo`, "handle": "foo:btn"
{% include '@bar:btn' with data %} // outputs `btn` from site `bar`, "handle": "bar:btn"
```

### DIFF
--- a/packages/core/test/utils.spec.js
+++ b/packages/core/test/utils.spec.js
@@ -298,6 +298,46 @@ describe('Utils', () => {
         it('returns correct path to file with extension', () => {
             expect(utils.relUrlPath('/path/to/image.png', '/path/b', opts)).toEqual('to/image.png');
         });
+
+        // for static builds
+        const opts2 = {
+            ext: '.html',
+            relativeToCurrentFolder: true,
+        };
+        const opts3 = {
+            relativeToCurrentFolder: true,
+        };
+        it('returns correct relative path from current directory for same directory', () => {
+            expect(utils.relUrlPath('/path/to/a', '/path/to/b', opts2)).toEqual('./a.html');
+        });
+
+        it('returns correct relative path from the current directory for parent directory', () => {
+            expect(utils.relUrlPath('/path/to/a', '/path/b', opts2)).toEqual('./to/a.html');
+        });
+
+        it('returns toPath if it is full url even if `relativeToCurrentFolder` is true', () => {
+            expect(utils.relUrlPath('https://fractal.build', '/path/b', opts2)).toEqual('https://fractal.build');
+        });
+
+        it('returns toPath with extension if it is already a relative path from current directory', () => {
+            expect(utils.relUrlPath('./path/to/a', '/path/b', opts2)).toEqual('./path/to/a.html');
+        });
+
+        it('returns toPath without extension if it is already a relative path from current directory', () => {
+            expect(utils.relUrlPath('./path/to/a', '/path/b', opts3)).toEqual('./path/to/a');
+        });
+
+        it('returns correct path to root from current directory', () => {
+            expect(utils.relUrlPath('/', '/path/b', opts3)).toEqual('./../..');
+        });
+
+        it('returns correct path to root with extension from current directory', () => {
+            expect(utils.relUrlPath('/', '/path/b', opts2)).toEqual('./../index.html');
+        });
+
+        it('returns correct path to file with extension from current directory', () => {
+            expect(utils.relUrlPath('/path/to/image.png', '/path/b', opts2)).toEqual('./to/image.png');
+        });
     });
 });
 
--- a/packages/core/test/utils.spec.js
+++ b/packages/core/test/utils.spec.js
@@ -328,11 +328,11 @@ describe('Utils', () => {
         });
 
         it('returns correct path to root from current directory', () => {
-            expect(utils.relUrlPath('/', '/path/b', opts3)).toEqual('./../..');
+            expect(utils.relUrlPath('/', '/path/b', opts3)).toEqual('../..');
         });
 
         it('returns correct path to root with extension from current directory', () => {
-            expect(utils.relUrlPath('/', '/path/b', opts2)).toEqual('./../index.html');
+            expect(utils.relUrlPath('/', '/path/b', opts2)).toEqual('../index.html');
         });
 
         it('returns correct path to file with extension from current directory', () => {
--- a/packages/core/test/utils.spec.js
+++ b/packages/core/test/utils.spec.js
@@ -299,6 +299,10 @@ describe('Utils', () => {
             expect(utils.relUrlPath('/path/to/image.png', '/path/b', opts)).toEqual('to/image.png');
         });
 
+        it('returns correct relative path to file with extension', () => {
+            expect(utils.relUrlPath('../to/image.png', '/path/b', opts)).toEqual('../to/image.png');
+        });
+
         // for static builds
         const opts2 = {
             ext: '.html',
@@ -337,6 +341,10 @@ describe('Utils', () => {
 
         it('returns correct path to file with extension from current directory', () => {
             expect(utils.relUrlPath('/path/to/image.png', '/path/b', opts2)).toEqual('./to/image.png');
+        });
+
+        it('returns correct relative path to file with extension from current directory', () => {
+            expect(utils.relUrlPath('../to/image.png', '/path/b', opts2)).toEqual('../to/image.png');
         });
     });
 });
--- a/packages/handlebars/src/helpers/path.js
+++ b/packages/handlebars/src/helpers/path.js
@@ -29,10 +29,9 @@ module.exports = function (fractal) {
             return path;
         }
 
-        return utils.relUrlPath(
-            path,
-            _.get(root._env.request || root._request, 'path', '/'),
-            fractal.web.get('builder.urls')
-        );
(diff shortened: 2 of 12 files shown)


packages/web/src/engine/globals/path.js
```diff
@@ -10,7 +10,7 @@ module.exports = function (app, engine) {
             req = req || this.lookup('request');
             return engine.env === 'server'
                 ? path
-                : utils.relUrlPath(path, _.get(req, 'path', '/'), app.web.get('builder.urls'));
+                : `./${ utils.relUrlPath(path, _.get(req, 'path', '/'), app.web.get('builder.urls')) }`;
```
mihkeleidast, line 13: what happens if the path returned from `utils.relUrlPath` is already relative? e.g. does `./../../some/path` behave the same as `../../some/path`?

mihkeleidast:
The adapters implement their own version of the path helper. Do you think those should also behave the same way? For example, Handlebars adapter does the same thing here: https://github.com/frctl/fractal/blob/main/packages/handlebars/src/helpers/path.js#L32

mihkeleidast:
> I can't lie, changing paths in the build of a project I'm not super familiar with gives me the fear. The tests pass, although I'm not sure anything tests following links in static builds.

The `utils.relUrlPath` is thoroughly covered with tests: https://github.com/frctl/fractal/blob/main/packages/core/test/utils.spec.js#L265

Maybe adding the logic inside that function would be better? Then it would also automatically cover the adapters as well, plus would be covered by tests (since adding it outside of that function is not covered by anything I think) and we could add more tests to validate your problem cases.

packages/web/src/engine/globals/path.js
13twelve: There is a [test](https://github.com/frctl/fractal/pull/1062/commits/12b63f69abdcf64918304a99464f723fcca01570#diff-4b766753284eb822a5a32a1b8e1f55480b64bdcae906a926dfc0b5cbfff955d1R322) for that case, it passes

13twelve:
Makes sense to move to the utils `relUrlPath` function. I added a `relativeToCurrentFolder` option incase of some future need to use/not use relative paths to the current folder. Passing this option through from the other adapters in the `/packages/` folder. Also added new tests to test for usage of this new option and also updated the snapshots for the updated build paths with `./` prefixing the paths.

packages/handlebars/src/helpers/path.js
```diff
@@ -29,10 +29,9 @@ module.exports = function (fractal) {
...
         }
 
-        return utils.relUrlPath(
-            path,
-            _.get(root._env.request || root._request, 'path', '/'),
-            fractal.web.get('builder.urls')
-        );
+        return utils.relUrlPath(path, _.get(root._env.request || root._request, 'path', '/'), {
+            ...fractal.web.get('builder.urls'),
+            relativeToCurrentFolder: true,
```
mihkeleidast, line 34: Regarding this - maybe it would be better to actually define this in the default config as such? Then it can be turned on/off in the project config, if need be (in case this actually breaks something for someone).

This would be the correct place: https://github.com/frctl/fractal/blob/main/packages/fractal/config.js#L117

packages/core/test/utils.spec.js
```diff
@@ -298,6 +298,46 @@ describe('Utils', () => {
...
+        it('returns toPath with extension if it is already a relative path from current directory', () => {
+            expect(utils.relUrlPath('./path/to/a', '/path/b', opts2)).toEqual('./path/to/a.html');
+        });
+
+        it('returns toPath without extension if it is already a relative path from current directory', () => {
+            expect(utils.relUrlPath('./path/to/a', '/path/b', opts3)).toEqual('./path/to/a');
+        });
+
+        it('returns correct path to root from current directory', () => {
+            expect(utils.relUrlPath('/', '/path/b', opts3)).toEqual('./../..');
```
mihkeleidast, line 331: I don't think this looks correct, maybe it should not touch the url if it is already relative (starts with `.`)

13twelve:
I'm now wondering if this `relUrlPath` should just be:

```JavaScript
relUrlPath(toPath, fromPath, opts) {
    return Path.extname(toPath).length > 0 || toPath.startsWith('http') || toPath.startsWith('.') || toPath.endsWith('/') ? toPath : toPath + (opts.ext || '');
}
```

If everything uses absolute paths and we just add the file extension if needed for static builds then links should always work.

I guess the use case for the relative paths was so that users could do a Fractal build and then double-click the generated index file and navigate around without a web server. 

But, I'm not sure this is needed. It mostly works but you can't flip between the various info panels for a pattern, at least in Chrome, because you hit `Uncaught DOMException: Blocked a frame with origin "null" from accessing a cross-origin frame.` errors as Chrome treats frames and iframes as different origins when browsing local files. So, I'd argue the static builds need a web server and so the Fractal navigation can use absolute paths instead of relative ones. I'd imagine most people upload the Fractal build to a web server to share with their teams.

Or, maybe it is still useful to be able to browse a Fractal pattern library locally without needing a web server and this just comes with the caveat that you can't swap the info panels when browsing a pattern (I guess the fix would be a page refresh?).

@mihkeleidast what do you think? 

As I'm keen to allow `:` in component handles I'll happily continue updating the method I've started, or switch to the simpler `relUrlPath` and update the tests accordingly.

mihkeleidast:
> I guess the use case for the relative paths was so that users could do a Fractal build and then double-click the generated index file and navigate around without a web server.

Not necessarily, the relative paths also allow users to upload the static build to any subdirectory on a server and have things just work. I use it daily in my work so let's not remove that functionality :)

13twelve:
> > I guess the use case for the relative paths was so that users could do a Fractal build and then double-click the generated index file and navigate around without a web server.
> 
> Not necessarily, the relative paths also allow users to upload the static build to any subdirectory on a server and have things just work. I use it daily in my work so let's not remove that functionality :)

Ah yes, fair point. I guess we could have a `base_path` type thing for the build in the config but then you might need to have different base paths for the build folder locally and on your server and now this is sounding like a pain.

I'll look at changes for your earlier comments 👍🏻

packages/handlebars/src/helpers/path.js
13twelve: Yes - much better - thanks

Updated in [1d4d1af](https://github.com/frctl/fractal/pull/1062/commits/1d4d1afffbae6fad5bfb91fd66f7371fb1cf9ad1)

packages/core/test/utils.spec.js
13twelve: Agreed - its a bit odd creating `./../` type urls.

Updated in [
ffafb4c](https://github.com/frctl/fractal/pull/1062/commits/ffafb4c6847cec5f5cbb2950cd73fcab7c51b5b9)

13twelve:
@mihkeleidast updates made - I prefer it with the config setting like this. I guess once done, the [Fractal.build Web UI Configuration page](https://fractal.build/guide/web/configuration-reference.html) needs a new entry

mihkeleidast:
@13twelve can you resolve the conflicts?

13twelve:
@mihkeleidast done 👍🏻

packages/core/test/utils.spec.js
```diff
@@ -299,8 +299,44 @@ describe('Utils', () => {
             expect(utils.relUrlPath('/path/to/image.png', '/path/b', opts)).toEqual('to/image.png');
         });
 
-        it('returns correct relative path to file with extension', () => {
-            expect(utils.relUrlPath('../to/image.png', '/path/b', opts)).toEqual('../to/image.png');
```
mihkeleidast, line 303: Do not remove this test, as it was just added with another PR.

mihkeleidast (changes requested):

packages/core/test/utils.spec.js
13twelve: Ack - the similar named tests confused me. Added it back, also added a variation with relativeToCurrentFolder being used.

mihkeleidast (approved):

13twelve:
@mihkeleidast thanks for your time and patience! 🙌🏻

mihkeleidast:
Cheers, thanks for working on this!

PR #1065, opened 2021-07-14
fix: return correct relative path if preexisting file ext

### LINKED ISSUE #1064: Reletive paths to static resouces with file extensions get .html appended to them
When using the [`path` handlebars helper](https://fractal.build/guide/core-concepts/view-templates.html#path) to generate relative paths for assets all of the assets are appended with `.html`, even though they have an extension already.

E.g. `pathTo/image.jpg` is being output as `pathTo/image.jpg.html`.

I did some digging into the code and found out this is happening in the `relUrlPath` function within the `core/src/utils.js` file.  Some sluthing on GitHub led me to #575.  This change doesn't verify there are existing extensions and adds the `.html` extension irregardless of the original path name.

It looks like one option is to [override this the default extension](https://fractal.build/guide/web/configuration-reference.html#builder-urls-ext) by adding `fractal.web.set('builder.urls.ext', null);` to one's `fractal.js` configuration file. 

Note: [Relevant message on Discord](https://discord.com/channels/664450585168314409/664450890660446229/864603137474166815).

1. In a component markdown file use the `path` helper in a handlebars template.
    ```
    {{path '../.../markdown-assets/documentation/image.jpg'}}
    ```
2. Inspect the rendered output HTML, the output is `../.../markdown-assets/documentation/image.jpg.html`

**Reproduces how often:**

`100%`

### Context

-   Fractal version: 1.5.8
-   Node version: v14.17.1
-   OS: Linux Mint 20.2 (Ubuntu 20.04), macOS 11.4

ScottMorris:
This PR closes #1064.

It addresses the issue by using regex to find a file extension on the `toPath` when the path is found to start with a `.`.  The solution also incorporates a test to handle this edge case.

### DIFF
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -140,7 +140,7 @@ module.exports = {
         const ext = opts.ext || '';
 
         if (toPath.startsWith('.')) {
-            return ext === '' ? toPath : toPath + ext;
+            return Path.extname(toPath) || ext === '' ? toPath : toPath + ext;
         }
 
         fromPath = getStaticPagePath(fromPath).replace(/\\/g, '/');
--- a/packages/core/test/utils.spec.js
+++ b/packages/core/test/utils.spec.js
@@ -298,6 +298,10 @@ describe('Utils', () => {
         it('returns correct path to file with extension', () => {
             expect(utils.relUrlPath('/path/to/image.png', '/path/b', opts)).toEqual('to/image.png');
         });
+
+        it('returns correct relative path to file with extension', () => {
+            expect(utils.relUrlPath('../to/image.png', '/path/b', opts)).toEqual('../to/image.png');
+        });
     });
 });
 


packages/core/src/utils.js
```diff
@@ -140,7 +140,8 @@ module.exports = {
         const ext = opts.ext || '';
 
         if (toPath.startsWith('.')) {
-            return ext === '' ? toPath : toPath + ext;
+            const toPathExt = /^.+\.([^.]+)$/.exec(toPath);
```
mihkeleidast, line 143: Use `Path.extname(toPath)` as is used a few lines below - `Path.extname(toPath) ? toPath : toPath + ext`.

mihkeleidast (changes requested):

packages/core/src/utils.js
ScottMorris: D'oh, good suggestion!

ScottMorris: I have mode this update.

mihkeleidast (approved):

mihkeleidast:
Very nice, thank you!

PR #1067, opened 2021-07-20
fix: relative path generated to self

ScottMorris:
- The path generated to self will be emitted as ./self.ext
- This fixes an issue introduced by #1062


mihkeleidast (approved):
Looks great, thanks! Going to release this ASAP.

PR #1081, opened 2021-09-24
fix: don't lowercase name before titlizing to generate label

khamer:
This is a minor change that, when generating labels for components, uses the original case version of `name` instead of the lowercased version. This means having component folder names like `someComponent/` correctly become `Some Component`, whereas without this change, it becomes `Somecomponent`.

### DIFF
--- a/packages/core/src/mixins/entity.js
+++ b/packages/core/src/mixins/entity.js
@@ -34,7 +34,7 @@ module.exports = mixin(
                 this._app = parent.source._app;
                 this.name = utils.slugify(name.toLowerCase());
                 this.handle = this._handle(config);
-                this.label = config.label || this._label(config);
+                this.label = config.label || this._label(name);
                 this.title = config.title || this._title(config);
                 this.order = _.isNaN(parseInt(config.order, 10)) ? 10000 : parseInt(config.order, 10);
                 this._isHidden = config.isHidden || config.hidden || false;
@@ -81,8 +81,8 @@ module.exports = mixin(
                 return this.name;
             }
 
-            _label() {
-                return utils.titlize(this.name);
+            _label(name) {
+                return utils.titlize(name);
             }
 
             _title() {
--- a/packages/fractal/src/api/docs/doc.js
+++ b/packages/fractal/src/api/docs/doc.js
@@ -23,8 +23,8 @@ module.exports = class Doc extends Entity {
         return this.name === 'index';
     }
 
-    _label(config) {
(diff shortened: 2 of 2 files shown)


packages/core/src/mixins/entity.js
```diff
@@ -34,7 +34,7 @@ module.exports = mixin(
                 this._app = parent.source._app;
                 this.name = utils.slugify(name.toLowerCase());
                 this.handle = this._handle(config);
-                this.label = config.label || this._label(config);
+                this.label = config.label || this._label(name);
```
mihkeleidast, line 37: The [Doc class](https://github.com/frctl/fractal/blob/main/packages/fractal/src/api/docs/doc.js#L26) expects `_label` method be passed the config object, not name.

mihkeleidast (changes requested):
I'm not sure about this change at all. `name` is [documented](https://fractal.build/guide/components/configuration-reference.html#name) as such: "values must be all lowercase, and contain only alphanumeric characters with hyphens or underscores for word seperators". 

Is there a reason for using camelCase instead of kebab-case for naming files/folders?

khamer:
Yes - the files themselves are also being referenced by the CMS (Craft) by slug, and those slugs need to be alphanumeric identifiers (no dashes.)

I will update doc.js too, I don't see anything in the Doc class's _label method that would require access to the whole config object.

Maybe a better question is - what does this hurt? I believe it just improves the the label generation, but doesn't change what files Fractal looks at, handles, etc.

mihkeleidast (approved):

mihkeleidast:
Thanks, looks good! I'll try to get this merged once GitHub Actions starts working properly again.

mihkeleidast:
@khamer I have no idea why GitHub doesn't allow me to run the workflows or merge main in to fix the out-of-dateness. Would you mind filing the PR again? Thanks.

khamer:
@mihkeleidast Sure, but if you'd like, check this one once more - I pulled all the subsequent upstream commits to my fork, so it might let you merge cleanly now. If not, I'll just make a new PR.

mihkeleidast:
Oh cool, the "Approve and Run" button appeared now!

mihkeleidast:
@khamer do you have any idea why the diff isn't clean? 🤔

khamer:
Hrm, I'm not, but reset my main branch to upstream and cherrypicked just my two commits and now the diff looks clean, and files changed on github shows the right two instead of twelve.

mihkeleidast:
Thanks, looks good now!

PR #1085, opened 2021-10-11
Fix typo in Nunjucks readme

querkmachine:
The code references .njk files but the associated code comment references .nunj files. 

Although there is no formal standard for Nunjuck file extensions, the community has largely opted to adopt .njk and it is the recommendation in the Nunjucks spec, so I assume that to be the "correct" option that is meant to be shown here.


mihkeleidast:
Yes, thanks!

PR #1100, opened 2021-11-04
feat: make navigation style configurable

### LINKED ISSUE #1091: Make subnavigation for variants configurable
In older versions of Mandelbrot the variants appeared beneth the link of the module. In recent versions a second navigation opens. 

Please make  this design decision configurable.
If a user wants to stick with the old version, newer versions of fractal cannot be installed because the design decisions seems to reflect into the core-code of fractal.

For the time being users of the older mandelbrot version have to stick to at last fractal 1.5.2.

mihkeleidast:
Resolves #1091.

This makes the new split navigation style opt-in,  which in retrospect is how we should've approached this from the start. Apologies on my part.


mihkeleidast:
Docs issue: https://github.com/frctl/fractal-docs/issues/107

PR #1129, opened 2022-01-19
build: Update marked from 2.0.3 to 4.0.10

Schleuse:
Updates marked`-dependency from `2.0.3` to `4.0.10` for some security fixes.

- https://github.com/advisories/GHSA-5v2h-r2cx-5xgj

Also Closes #1126

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16129,9 +16129,9 @@
       "dev": true
     },
     "marked": {
-      "version": "2.0.3",
-      "resolved": "https://registry.npmjs.org/marked/-/marked-2.0.3.tgz",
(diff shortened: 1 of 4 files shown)


mihkeleidast:
Marked 3 dropped support for Node 10, which we still do support. Not sure if we should update this before dropping Node 10 ourselves.

Schleuse:
Oh, I thought fractal only supported node versions which are not EOL (https://github.com/frctl/fractal#requirements). Node 10 already entered EOL on 2021-04-30

mihkeleidast:
True, that's our general requirement. But doing a major version release when dropping an older Node version support is essentially the open source standard, so we've been trying to align this major release with some other long-standing developments that would also be breaking. I'll ping the team to see where we are on those, as none of us can really prioritize these developments.

Schleuse:
Thanks @mihkeleidast for the clarification

PR #1131, opened 2022-01-20
feat: add create-fractal package

### LINKED ISSUE #635: Add npm initializer package

npm [supports custom initializers](https://docs.npmjs.com/cli/init) to ease up setuping a package. If we add a package named `@frctl/create-fractal`, that could be used via `npm init @frctl/fractal`. This would essentially allow users to use the `fractal new` command without installing the `@frctl/fractal` package globally.

Add a package named `@frctl/create-fractal` that does the same internal things that `fractal new` cli command does.

mihkeleidast:
Closes #635.


packages/create-fractal/README.md
```diff
@@ -0,0 +1,18 @@
+# @frctl/create-fractal
+
+New Fractal pattern library initializer.
+
+[![NPM Version](https://img.shields.io/npm/v/@frctl/create-fractal)](https://www.npmjs.com/package/@frctl/create-fractal)
+
+## Usage
+
+Run `npm init @frctl/fractal [dir]` in your terminal. `[dir]` is the new folder you want to create your project in.
```
mihkeleidast, line 9: We could also rename the package to just allow `npm init @frctl` or do something longer, e.g. `npm init @frctl/library`, though I'm not sure it would be better. Did not check if we could do just `npm init fractal`, e.g. if the unscoped package name is free on npm.

PR #1230, opened 2022-10-25
fix: only expand tree's with matches

Schleuse:
We're using fractal for pretty big stylesguides with 1700+ patterns which makes the search pretty much a necessity. 
As suggested in #1202, this PR makes it that trees that don't have any matching children, won't get expanded anymore.

**before:**

https://user-images.githubusercontent.com/2717384/197841929-74360cf5-2bfa-4f58-a014-376e44bfbe1b.mp4

**after:**

https://user-images.githubusercontent.com/2717384/197841953-7e88abe2-999e-4b11-b13b-8c241fb9fe27.mp4


lgollut:
Is there anything blocking the release of that fix ?

LeBenLeBen:
Moved to #1252 so it can be released soon

PR #1231, opened 2022-10-26
fix: replace Promise.props

Schleuse:
Recently Bluebird got replaced (7a2ed63fe87304a3c7a145aef3a9426c6857b27f) with the native Promise implementation, but some calls to blubird functionality were left behind (namely [`Promise.props`](http://bluebirdjs.com/docs/api/promise.props.html)).
Because of that, the examples weren't able to start anymore:

```
cd examples/handlebars
npm run start

@frctl/handlebars-example@0.3.11 start
> fractal start --sync

✘ TypeError: Promise.props is not a functionflatten Completed in 1ms
```

This PR tries to replace all calls to `Promise.props` with a compatible implementation that works with native Promises.


mattc-cogapp:
Thanks for this @Schleuse - I'll take a look this evening! I'll also take a look at the tests, as ideally this would have been picked up by the CI runner.

PR #1252, opened 2023-01-26
fix: only expand tree's with matches

LeBenLeBen:
Superseeding #1230 but without the recent cleanup so it can be released.


LeBenLeBen:
@Schleuse I tested your code and found an issue when there’s a match in the parent collection label, it was not showing the child items anymore:

image

I have adapted the code a bit to get this result:

SCR-20230126-dft

